PR #20, opened 2023-02-13
feat: Stack::<u256>

### LINKED ISSUE #16: feat: Stack
# Feature Request

**Describe the Feature Request**

Implement a Stack data structure in Cairo 1.0

**Describe Preferred Solution**

[EDIT] After discussing it, the stack should be implemented using a dictionary.
The stack should expose three "main" methods to the user : `new`, `pop` and `push`. On top of that, we can implement utility functions such as `len`, `peek` and `is_empty`.

**Describe Alternatives**

**Related Code**

```
struct Stack<T> {
    elements: Array::<T>, 
}

trait StackTrait<T> {
    fn new() -> Stack::<T>;
    fn push(ref self: Stack::<T>, value: T);
    fn pop(ref self: Stack::<T>) -> Option::<T>;
    fn peek(ref self: Stack::<T>) -> Option::<T>;
    fn len(ref self: Stack::<T>) -> usize;
    fn is_empty(ref self: Stack::<T>) -> bool;
}
```
**Additional Context**

While the Stack implementation seems pretty trivial, I have given a try at it and encountered two major issues that should be solved with the next release of Cairo 1.

- Variable was previously moved
Here's a snippet of code illustrating the problem
```
// Variable was previously moved
fn peek(ref self: Stack::<T>) -> Option::<T> {
        let mut elements = self.elements;
        let mut last_index = elements.len() - 1_usize;
        let last = elements.get(last_index);
        self = Stack { elements };
        last
}
```
Hypothesis: the borrow checker complains here because I took ownership of `self` in the operation `elements.len() - 1_usize`. Therefore, I can't reuse it later in the code. This would require generic type `T` to implement the `Dup` trait (explanation from @spapinistarkware) which is a feature that will soon be implemented.

- Variable not dropped.
A variable that is declared and not returned at the end of a function generates the error `Variable not dropped.` I guess the comment above for the `Dup` trait stands for the `Drop` trait as well.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

msaug:
Implements an array-based u256 stack
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

A functional array-based U256Stack

Issue Number: Fixes #16 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/src/data_structures/stack.cairo
+++ b/src/data_structures/stack.cairo
@@ -2,15 +2,22 @@
 //!
 //! # Example
 //! ```
-//! use quaireaux::data_structures::stack::StacTrait;
+//! use quaireaux::data_structures::stack::StackTrait;
 //!
 //! // Create a new stack instance.
 //! let mut stack = StackTrait::new();
-//! # TODO: Add example code for the different methods.
+//! // Create an item and push it to the stack.
+//! let mut item  = ArrayTrait::<u256>::new();
+//! item.append(1_u256);
+//! stack.push(item);
+//! Remove the item from the stack;
+//! stack.pop();
+//! ```
 
 // Core lib imports
 use array::ArrayTrait;
 use option::OptionTrait;
+use quaireaux::utils::array_slice;
 
 const ZERO_USIZE: usize = 0_usize;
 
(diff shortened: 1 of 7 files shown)


src/data_structures/stack.cairo
```diff
@@ -5,27 +5,94 @@
...
+        let length = elements.len();
+        self = Stack { elements };
+        length
     }
+
+    fn is_empty(ref self: Stack) -> bool {
+        self.len() == ZERO_USIZE
+    }
+}
+fn array_slice(
```
abdelhamidbakhta, line 72: nit: move this function in `array_utils`

src/data_structures/stack.cairo
```diff
@@ -5,27 +5,94 @@
...
 impl StackImpl of StackTrait {
-    /// Create a new stack instance.
     #[inline(always)]
     fn new() -> Stack {
-        Stack {}
+        let mut elements = ArrayTrait::<Array::<u256>>::new();
+        Stack { elements }
+    }
+
+    fn push(ref self: Stack, value: Array::<u256>) {
```
abdelhamidbakhta, line 38: please add Rust style documentation on all functions

abdelhamidbakhta (approved):
looks good to me when tests pass and problems of traits derivation are fixed

msaug:
`peek` now works correctly. The only problem remaining is the `Value moved` error in the `pop` function. I can't figure out how to fix this.

msaug:
Everything should work now @abdelhamidbakhta, can you give it a second look because a lot of things have changed since your first approval.

But for some reason it doesn't work with the CI 😭 

For now the `pop` function doesn't take `Stack` as `ref` but instead as a `mut` variable and it is returned explicitly. I don't know why it doesn't work with ref - there must be secret shenanigans that I don't understand yet

PR #54, opened 2023-03-11
Addin cron on tests

gaetbout:
Adding CRON on tests workflow to run them once a day to try and keep the repo up to date with latest

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,11 @@
 name: test
 
-on: [push, pull_request]
+on: 
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    - cron: "0 8 * * *"
 
 env:
   CAIRO_COMPILER_VERSION: 1.0.0-alpha.3


abdelhamidbakhta:
> Adding CRON on tests workflow to run them once a day to try and keep the repo up to date with latest

The problem is that you have not included the upgrade of latest Cairo version, we need to think about how to enable this. Meanwhile let's put the PR as draft

gaetbout:
We could clone Cairo1's repo, build the binaries in the CI. 
On top of that we could use the cache to only make an incremental update?

abdelhamidbakhta:
> We could clone Cairo1's repo, build the binaries in the CI. On top of that we could use the cache to only make an incremental update?

Good idea, do you want to try this approach ?

gaetbout:
Agree, makes sense as this CI will prob be longer to run

abdelhamidbakhta:
> Agree, makes sense as this CI will prob be longer to run

Yep exactly

gaetbout:
Running into some issues due to the fact that there is no Cargo.toml at the root level of the project.
Looking into solving it.

gaetbout:
@abdelhamidbakhta 
Should be working now, but as the repo uses an "old" version of Cairo maybe it is best not to merge it for now.
Unless someone updates the repo to the latest version

gaetbout:
As we'll stick to keep the repo as close as possible to the latest official release, this isn't useful anymore

PR #67, opened 2023-04-01
Refactor to their own crate

gaetbout:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Re-imported check_gas to avoid utils::utils::check_gas;
Moved crate related fn into their own crate file. 
Optimized import

Maybe instead of creating a file called by the name of the folder, I should put those fn in the lib.cairo file?

**REMOVED:** Renamed data-structures to data-structure (singular) to match with other crates


quaireaux/data_structure/cairo_project.toml
```diff
@@ -1,3 +1,3 @@
 [crate_roots]
-data_structures = "src"
+data_structure = "src"
```
enitrat, line 2: why no 's'? There are multiple DS

enitrat (commented):
I don't understand the reason for this refactoring. What does it bring? I don't see any advantages with this new structure

edit: I must've missed a few PR's but I am talking about the organization in sub-crates

quaireaux/data_structure/cairo_project.toml
gaetbout: Then we should put maths and sortings just feels weird that one in is plural while all the others are singular

gaetbout:
> does it bring? I don't see any advantages with this new structure
> 
> edit: I must've missed a few PR's but I am talking about th

The utils crate was cluttered with stuff only useful in very specific case. 
So, for example, I moved some stiff to a math file in the math crate. 
This PR is the same but for the two other crates.but I understand that these changes aren't that useful. (besides the re-import of check_gas)

quaireaux/data_structure/cairo_project.toml
enitrat: No, because  the "data-structure**s**" crate defines multiple data **structures**, and pluralizing the name helps convey that there is a collection of them in the crate.

However in the case of "sorting", it's a specific concept that refers to the process of sorting items. While you could use a plural form like "sorting-algorithms" or "sorting-methods", but I think the singular form conveys the idea more efficiently.

just a nitpick but I think it was worth sharing the thought :)

gaetbout: I'll revert this change then 😄

gaetbout: I wasn't sure about that change, so your concern confirms my doubts 😃

abdelhamidbakhta (approved):
Lgtm

PR #86, opened 2023-05-03
Feat: add number -> ascii functions

wraitii:
Sets a simple layout for a string library, and add functions to print felts and other values as ASCII strings.

Open question: should this be a trait, should the base be a parameter.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- [ ] Yes
- [x ] No


gaetbout:
Could be handy if it would follow some rust standards such as:
(Making in an interface and impl could already help a lot)
```rust
String::new();
String::from("Hello, world!");
myString.len();
myString.capacity();
```
https://doc.rust-lang.org/std/string/struct.String.html

wraitii:
Added a simple string type wrapping `Array<u8>`.
I'm experimenting with providing generic implementations in a submodule for explicitness, need to test something first.

Note that I'm basically providing just `StringTrait::from_number(1234) // prints '1234'`

wraitii:
> I really like this and WDYT of having a `StorageString` type that would use felts when stored in a contract ?

Not entirely sure I get what you mean, wouldn't it be better to have a StorageAccess implementation that uses bit-packing for that instead ?

wraitii:
Yes we're talking about the same thing.

I didn't really want to spend too much more time on this as of yet, as my use-case for briq was just the 'raw felt into a string' thing, so if it's OK to merge this early iteration I'll fix the format check and we can merge this.

LucasLvy:
> I didn't really want to spend too much more time on this
yeah no worries i was just throwing an idea that emerged from my gigantic brain. It'll be done later

wraitii:
> > I didn't really want to spend too much more time on this
> 
> yeah no worries i was just throwing an idea that emerged from my gigantic brain. It'll be done later

TBH On second thought I do somewhat agree with you, we probably want a 'PackedString' type that's more storage efficient. We could have a 'FlexibleString' type on top that enums on regular string / packed string.
I'll add a little more code and update the PR.

gaetbout:
I'm fine doing the packing once this is merge 😄 
I'm just worried about the extra cost of doing (potentially) a lot of AND and OR...

wraitii:
Proposal updated to an absurd degree:
- I implement a simpler StringTrait, with parametrisation on the container and codepoint
- I implement this type on Array<u8>, which can be seen as a string of ascii
- I implement a separate 'ShortString' wrapper on felt252

This lays down the foundation for:
- UTF-8 aware string, which would have a different code-point type.
- packed string, which would not be an array underneath.

Note that I don't implement `len`, for the simple reason that `StringTrait` should imply `ArrayTrait`, but that's not possible until `ArrayTrait` is updated to take <T, U> and be more of an `ArrayLike` trait.
According to Shahar, this is doable once the Cycle issue (https://github.com/starkware-libs/cairo/issues/3031) is fixed

wraitii:
Ok, better update IMO:
- There is no string trait
- There is a 'sequence' trait, implemented by all string-like types
- Array<u8> is defined as an ASCII string
- ShortString is a wrapper around the felt252-based short string

LucasLvy:
You don't like formatting do you ? 🔫

quaireaux/string/src/short_string.cairo
```diff
@@ -0,0 +1,92 @@
...
+impl ShortStringConstructor of Construct<ShortString> {
+    fn new() -> ShortString {
+        ShortString { data: 0 }
+    }
+}
+
+use debug::PrintTrait;
+
+// Implement a 0-terminated (if size < 31) wrapper around the short-string native to Cairo 1.
+impl ShortStringImpl of SequenceTrait<ShortString, u8> {
```
LucasLvy, line 32: WDYT of that ?
```suggestion
impl String of SequenceTrait<ShortString, u8> {
```

wraitii: No longer applicable

quaireaux/string/src/utils.cairo
```diff
@@ -0,0 +1,12 @@
+// These traits are intended to be temporary until similar traits are in the Cairo standard library,
+// thus their placement in utils.
+
+trait RandomAccessTrait<T, U> {
+    fn get(self: @T, index: usize) -> Option<Box<@U>>;
+    fn at(self: @T, index: usize) -> @U;
+}
```
LucasLvy, line 7: We now have the index trait. It's implemented for vec  https://github.com/keep-starknet-strange/alexandria/blob/fc1de0371d7ee0e9fa61c4c156df737a2adbdb04/alexandria/data_structures/src/vec.cairo#L118-L123

LucasLvy (changes requested):
minor change plz and don't forget to format 💋

gaetbout:
Not stale, we are waiting for a stable version of compiler v2

PR #93, opened 2023-05-17
base64

neotheprogramist:


quaireaux/math/src/base64.cairo
```diff
@@ -0,0 +1,225 @@
+mod Decoder {
```
tarrencev, line 1: I don't think this should be in the `math` crate. Perhaps create an `encoding` crate?

tarrencev (commented):
I think an ideal interface for this would be to define traits like

```
trait Base64Encoder<T> {
    fn encode_base64(self: T) -> Array<u8>;
}
```

and it can be implemented for different types

quaireaux/math/src/base64.cairo
abdelhamidbakhta: I agree with @tarrencev

PR #98, opened 2023-05-22
base64

neotheprogramist:


tarrencev (approved):
looks great

LucasLvy:
How about a generic encoder trait and then you implement it for base64 ?

neotheprogramist:
@LucasLvy Could you take a look at the code now?
If you have any further suggestions lmk :)

gaetbout:
Could you update the main Readme ## Features section?

alexandria/encoding/src/base64.cairo
```diff
@@ -0,0 +1,326 @@
...
+            if *data[data.len() - 2] == '=' {
+                p += 1;
+            }
+            decode_loop(p, ref data, 0, ref result);
+        }
+        result
+    }
+}
+
+impl u8Base64UrlDecoder of Decoder<u8> {
```
gaetbout, line 73: As both decoder are the same, I wouldn't do two just for the sake of having one that suit each name.
But if you think it makes sense I guess we can keep it, no strong opinion

alexandria/encoding/src/math.cairo
```diff
@@ -0,0 +1,20 @@
...
+    }
+}
+
+fn shl(x: u128, n: u128) -> u128 {
+    x * fpow(2, n)
+}
+
+fn shr(x: u128, n: u128) -> u128 {
+    x / fpow(2, n)
+}
```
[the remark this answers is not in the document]
neotheprogramist: This is the import problem I was writing about in the group.
I don't know how to import functions between crates correctly, so I left that doubled but working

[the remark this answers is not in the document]
gaetbout: In the encoding scarb.toml put
```toml
[dependencies]
quaireaux_math = { path = "../math/" }
```
in the encoding cairo_project.toml put:
```
quaireaux_math = "../math/src"
```
Move all to the math file and update local import of sha256
then in encoding crate you can use math fn like:
```
use quaireaux_math::{count_digits_of_base, pow};
```

gaetbout:
calling @LucasLvy for final review

alexandria/encoding/cairo_project.toml
```diff
@@ -0,0 +1,3 @@
+[crate_roots]
+alexandria_encoding = "src"
+alexandria_math = "../math/src"
```
LucasLvy, line 3: Does it need to be here ?

alexandria/encoding/src/base64.cairo
```diff
@@ -0,0 +1,246 @@
+use core::clone::Clone;
+use array::ArrayTrait;
+use integer::{upcast, downcast};
+use option::OptionTrait;
+use debug::PrintTrait;
+use alexandria_math::math::{shl, shr, U6_MAX, U8_MAX};
+
+trait Encoder<T> {
+    fn encode(data: Array<T>) -> Array<u8>;
```
LucasLvy, line 9:
```suggestion
    fn encode(data: T) -> Array<u8>;
```
How about this ? so you can encode anything ?

alexandria/encoding/src/base64.cairo
```diff
@@ -0,0 +1,246 @@
...
+}
+
+trait Decoder<T> {
+    fn decode(data: Array<T>) -> Array<u8>;
```
LucasLvy, line 13:
```suggestion
    fn decode(data: T) -> Array<u8>;
```
How about this ? so you can decode anything ?

alexandria/encoding/src/base64.cairo
```diff
@@ -0,0 +1,246 @@
...
+        }
+        if *data[data.len() - 2] == '=' {
+            p += 1;
+        }
+        decode_loop(p, ref data, 0, ref result);
+    }
+    result
+}
+
+fn decode_loop(p: u8, ref data: Array<u8>, d: usize, ref result: Array<u8>) {
```
LucasLvy, line 81: you can use a loop there (maybe you prefer recursive ?)

alexandria/encoding/src/base64.cairo
```diff
@@ -0,0 +1,246 @@
...
+    } else if (x == '_') {
+        63
+    } else if (x <= 'z') {
+        (x - 'a') + 26
+    } else {
+        0
+    }
+}
+
+fn encode_loop(
```
LucasLvy, line 141: you can use a loop there (maybe you prefer recursive ?)

LucasLvy (changes requested):

alexandria/encoding/cairo_project.toml
gaetbout: It is useful as it uses some code from math such as pow and shl and shr

alexandria/encoding/src/base64.cairo
neotheprogramist: With the current state of Cairo, I prefer recursion

Issue #95, opened 2023-05-18
fast_power fails for `2**127`


moodysalem:
https://github.com/keep-starknet-strange/alexandria/blob/922ea0d124ddb48f796cce1419dd56b8ba87740c/quaireaux/math/src/fast_power.cairo#L11-L29

This is because the final base*base multiplication overflows

Can use u256_mul_overflow

edisontim:
Are you working on this? If not I wouldn't mind having a look :)

moodysalem:
I'm not, just found a way to not use pow

lambda-0x:
@moodysalem this should be fixed now, can you confirm and close the issue?

LucasLvy:
Fixed by #104

PR #104, opened 2023-06-04
fix: integer overflow in fast_power

lambda-0x:
## Pull Request type

- [x] Bugfix

Issue Number: #95 

## What is the new behavior?

- panic with overflow error for certain inputs

- [ ] Yes
- [x] No

## Other information

I have choose to change all the parameter types to `u256` instead of using [`u128_wide_mul`](https://github.com/starkware-libs/cairo/blob/77ae28f996c0960ce5cfc926703f60bae8d5db5a/corelib/src/integer.cairo#L66C1-L69).

Because in the loop:
- we would need to do it in 2 places, so 2 type conversions to u256
-  `%` is only implemented for same types we cannot calculate `u256 % u128`, so would need to convert `modulus` to u256 
- after calculating modulus would needs to cast it back to u128

Also for some reason comment mentioned result of function is `felt252` instead of `u128`, so have fixed that as well.

### DIFF
--- a/alexandria/math/src/fast_power.cairo
+++ b/alexandria/math/src/fast_power.cairo
@@ -22,7 +22,7 @@ fn fast_power(mut base: u128, mut power: u128, modulus: u128) -> u128 {
     // TODO: Simplify thise conversions after https://github.com/starkware-libs/cairo/pull/3293 is merged and released.
     let mut base: u256 = u256 { low: base, high: 0 };
     let modulus: u256 = u256 { low: modulus, high: 0 };
-    let mut result: u256 = 1;
+    let mut result: u256 = u256 { low: 1, high: 0 };
 
     let res = loop {
         if power == 0 {
@@ -42,6 +42,6 @@ fn fast_power(mut base: u128, mut power: u128, modulus: u128) -> u128 {
     if high != 0 {
         panic_with_felt252('value cant be larger than u128')
     }
-    
+
     return low;
 }
--- a/alexandria/math/src/fast_power.cairo
+++ b/alexandria/math/src/fast_power.cairo
@@ -1,21 +1,30 @@
 //! # Fast power algorithm
 use array::ArrayTrait;
+use integer::{u128_wide_mul};
+use core::option::OptionTrait;
+use core::traits::TryInto;
 
-// Calculate the (base^power)mod modulus using the fast powering algorithm
-// # Arguments
-// * `base` - The base of the exponentiation
-// * `power` - The power of the exponentiation
-// * `modulus` - The modulus used in the calculation
-// # Returns
-// * `felt252` - The result of (base^power)mod modulus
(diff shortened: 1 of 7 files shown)


LucasLvy:
Make sure you use the same compiler version as the ci plz

LucasLvy:
also i'd suggest we keep the math in u128. If you want to create something with u256 plz create another function @gaetbout  agree with that ?

lambda-0x:
> Make sure you use the same compiler version as the ci plz

done

> also i'd suggest we keep the math in u128. If you want to create something with u256 plz create another function @gaetbout agree with that ?

@LucasLvy  as mentioned PR's comment, afaik there are only 2 ways to do this correctly, both of which involve `u256`.
1. either use `u256` (as i did)
2. or use [`u128_wide_mul`](https://github.com/starkware-libs/cairo/blob/77ae28f996c0960ce5cfc926703f60bae8d5db5a/corelib/src/integer.cairo#L66C1-L69) which would require:
    - we would need to do it in 2 places where we are doing multiplication, and to calculate modulo we would need to convert `(u128, u128)` to `u256`
    - `%` is only implemented for same types we cannot calculate `u256 % u128`, so would need to convert `modulus` to `u256` as well
    - after calculating modulus would needs to cast it back to `result` as u128 with `try_into` so would need an `unwrap`

so i think its would be much easier to read the code with current method, and also I think it would even be more efficient to just do cast once at beginning and once at end instead of doing it in the loop.

alexandria/math/src/fast_power.cairo
```diff
@@ -1,29 +1,47 @@
 //! # Fast power algorithm
 use array::ArrayTrait;
+use integer::{u128_wide_mul};
```
gaetbout, line 3:
```suggestion
use integer::u128_wide_mul;
```

LucasLvy:
Still not sure to understand what you mean but keep this function as u128 if it overflows it's the user's problem. If you want to create a function that uses u256 you can do it but leave this one as is plz

lambda-0x:
@LucasLvy please have a look at the linked issue, And consider this test case:

```cairo
assert(
    fast_power(
        2, 127, 340282366920938463463374607431768211454
    ) == 170141183460469231731687303715884105728,
    'invalid result'
);
```

The current implementation is broken because at some point it will try to multiple 2 very big numbers that cannot be stored in `u128`  and it will panic

> if it overflows it's the user's problem

no its not, the functions accepts `u128` and should work on for all values of type `u128`.

> If you want to create a function that uses u256

I don't want to create a function that uses `u256`, i want this function to return a valid result for a valid input.

---

I would love to fix this issue without using `u256` but afaik there isn't a way to do it (maybe now my above comment makes sense?). If you know a way please let me know.

LucasLvy (approved):

Issue #123, opened 2023-06-26
feat: numerical analysis and linear algebra libs


Bal7hazar:
# Feature Request

**Description**

Working on curves data is quite common for some protocols, having some numerical analysis methods available in a public library such as Alexandria would make sense to benefit all starknet buidlers.

The minimal scope I will need for my purpose is:
- interpolation with some options about extrapolation methods
- cumsum

But I will be glad to build a more relevant minimal scope that we could define together.

**Describe Preferred Solution**

The ability to add aleandria as a dep in my external project and use numerical methods such as:
- gradient
- trapezoidal rule
- ...

Or linear algebra functions:
- dot
- vdot
- eigen values
- ...

**Describe Alternatives**

The alternative would be to have this public library outside of Alexandria or worst, having a new implement in each protocol which need this.

**Related Code**

There are a lot of stuff here:
- https://numpy.org/devdocs/reference/routines.linalg.html
- https://docs.scipy.org/doc/scipy/reference/index.html#scipy-api

**Additional Context**

We should consider about which folder will handle which scope also (Math? Algebra? Numerical?)

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

PR #124, opened 2023-06-26
✨ Implement numerical and linalg features

Bal7hazar:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: #123

## What is the new behavior?

- Linear algebra

  - Dot product

- Numerical analysis

  - Trapezoidal rule
  - Cumsum a sequence
  - Diff a sequence
  - Interpolate y at x with different interpolation and extrapolation methods

- [ ] Yes
- [x] No

## Other information

Developments done with:
```bash
scarb 0.5.0-alpha.4 (6484cf7b6 2023-06-29)
cairo: 2.0.0-rc6 (https://crates.io/crates/cairo-lang-compiler/2.0.0-rc6)
```

Workload: ~ 6 hours


Bal7hazar:
@LucasLvy I'm done for a first review, I create 2 new folders (`numeric` and `linalg`) because they respectively have high growth potential, not sure if you prefered to have it in math.

alexandria/numeric/src/interp.cairo
```diff
@@ -0,0 +1,89 @@
...
+            Extrapolation::Constant(()) => *ys.at(xs.len() - 1_usize),
+        };
+        return y;
+    }
+
+    // [Compute] Interpolation
+    let mut index = 0_usize;
+    loop {
+        if index + 1_usize == xs.len() {
+            assert(0 != 1, 'Unexpected error');
```
LucasLvy, line 58:
```suggestion
            assert(0 == 1, 'Unexpected error');
```
You mean that right ?

alexandria/numeric/src/interp.cairo
```diff
@@ -0,0 +1,89 @@
...
+        }
+        assert(*xs.at(index + 1_usize) > *xs.at(index), 'Abscissa must be sorted');
```
LucasLvy, line 60: you should be able to use that:
```suggestion
        assert(*xs[index + 1_usize] > *xs.at(index), 'Abscissa must be sorted');
```

LucasLvy (changes requested):
Looks good but switch to rust doc (`///` instead of `//` above functions). And please use more explicit function names

alexandria/numeric/src/interp.cairo
Bal7hazar: Thank you, I'll do that

alexandria/numeric/src/interp.cairo
Bal7hazar: I replaced it by `false`  the I just removed it since this case cannot be reached logically.

Bal7hazar:
> Looks good but switch to rust doc (/// instead of // above functions). And please use more explicit function names
I replaced `interp`  by `interpolate` and `trapz`  to `trapezoidal_rule` which break from numpy naming, but it's ok I think.

Also `dot`, `cumsum`  and `diff` sounds to be explicit enough to me (it is named like this in other scientific libs I think). But I can change them also if it matters.

LucasLvy (approved):

Issue #130, opened 2023-07-04
bug: New interdependency between math and data_structures due to ed25519 feature


Bal7hazar:
# Bug Report

**Alexandria version:**

`main`

**Current behavior:**

Since the merge of ed25519 feature, there is a dependency between alexandria_math and alexandria_data_structures which force a user to add the alexandria_data_structures if he wants to only use alexandria_math.

**Expected behavior:**

I expected that each library would not be interdependent, but I could be wrong.

**Steps to reproduce:**

```bash
scarb add alexandria_math --git https://github.com/keep-starknet-strange/alexandria
scarb build
```

Generates the following error:

```bash
error: Identifier not found.
 --> ed25519.cairo:9:5
use alexandria_data_structures::array_ext::{ArrayTraitExt};
    ^************************^

error: Method `concat` not found on type "@core::array::Array::<core::integer::u8>". Did you import the correct trait and impl?
 --> ed25519.cairo:303:49
    let k: Array<u8> = sha512(r_string.snapshot.concat(pub_key.snapshot).concat(msg.snapshot));
                                                ^****^

error: Method `concat` not found on type "<missing>". Did you import the correct trait and impl?
 --> ed25519.cairo:303:74
    let k: Array<u8> = sha512(r_string.snapshot.concat(pub_key.snapshot).concat(msg.snapshot));
                                                                         ^****^
```

To solve this we could either not use alexandria_data_structures in alexandria_math, or merge both libraries.

Finally, this is a good opportunity to statuate on this situation and if possible add a test to prevent it in the future if we choose to avoid this situation.

**Related code:**

https://github.com/keep-starknet-strange/alexandria/blob/3fae57a07ba8904f8e8a4299448a5c2f443f8330/alexandria/math/src/ed25519.cairo#L9C1-L9C60

edisontim:
I forgot to add alexandria_data_structures to the scarb.toml, but why not have interdependencies? The only issue I can see is recursive dependency, is there something else that might be annoying?
The reason I ask is because this happens as well in the package alexandria_encoding with base64, which imports rhs and lsh from alexandria_math.

Bal7hazar:
> but why not have interdependencies?

I don't have strong opinion about interdependencies, I didn't notice any other so I was thinking that we have to avoid that. But as you mentioned base64 already did that so adding the dep in the scarb should be the right thing to do then.

[...]

PR #136, opened 2023-07-06
refactor: group packages

edisontim:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes

Library is currently divided in multiple scarb packages. 
Related to discussion on issue #130 

## What is the new behavior?

- Group all packages into one
- Remove circular dependencies

- [X] Yes
- [ ] No

As @gaetbout suggested we can create a tag that people can use to reference this repo "pre-package merge"


README.md
```diff
@@ -72,10 +73,10 @@ This repository is divided in 5 main packages:
 
 ### Installation
 
-Alexandria is a collection of [Scarb] packages, which can be installed by adding the following line to your `Scarb.toml`:
+Alexandria is a collection of cairo utilities, which can be installed by adding the following line to your `Scarb.toml`:
 ```toml
 [dependencies]
-alexandria_math = { git = "https://github.com/keep-starknet-strange/alexandria.git" }
+alexandria = { git = "https://github.com/keep-starknet-strange/alexandria.git" }
 ```
```
gaetbout, line 80: Could you put here an example on how it should be imported in a Cairo file?
I know it wasn't there before but this could be useful 😄

edisontim: Yes ofc :)

edisontim:
Also wondering if I should put 
```
[[target.starknet-contract]]
``` 
and 
```
[dependencies]
starknet = ">=2.0.0-rc0"
```
in the scarb.toml. Anyone familiar with scarb knows about this?

edisontim:
Also worried about the tests, they take a long time when running contract and non-contract cairo at the same time which is super weird. I'll have a look at that or ways to make it quicker

Bal7hazar:
> Also wondering if I should put
> `[[target.starknet-contract]]`
> and
> `[dependencies]`
> `starknet = ">=2.0.0-rc0"`

If `scarb build` and `scarb test` work fine without I assume we can remove them.
But I think some tests are actually using starknet features.

edisontim:
> If `scarb build` and `scarb test` work fine without I assume we can remove them. But I think some tests are actually using starknet features.

`scarb test` wasn't passing without so I added them :)

edisontim:
Before merging I would like to try to use alexandria in this form as a scarb package, I will see if it is possible via the tag or rev of scarb dependencies, considering it's a PR and not yet merged to main

Scarb.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "alexandria"
+version = "0.1.0"
+description = "Starknet utilities"
+homepage = "https://github.com/keep-starknet-strange/alexandria/"
+
+[[target.starknet-contract]]
+
+[dependencies]
+starknet = ">=2.0.0-rc0"
```
Bal7hazar, line 10: What about to set `starknet = ">=2.0.0"` or even `starknet = ">=2.0.1"` since release candidates led to build the 2.0.0 final version?

edisontim:
Was able to test this in a local project, PR is good to go

README.md
```diff
@@ -72,14 +75,20 @@ This repository is divided in 5 main packages:
...
 
-## Usage
+then add the following line in your `.cairo` file
 
+```rust
+use alexandria::math::sha512::{sha512};
```
LucasLvy, line 88:
```suggestion
use alexandria::math::sha512::sha512;
```

src/math/ed25519.cairo
```diff
@@ -222,6 +221,35 @@ impl PointIntoExtendedHomogeneousPoint of Into<Point, ExtendedHomogeneousPoint>
     }
 }
 
+/// Function that concats two arrays together
+/// # Arguments
+/// * `a` - First array.
+/// * `b` - Second array.
+/// # Returns
+/// * `Array<u8>` - true if the signature fits to the message and the public key, false otherwise.
+fn concat(a: @Array<u8>, b: @Array<u8>) -> Array<u8> {
```
LucasLvy, line 230: Weird to have this function here

LucasLvy (changes requested):

src/math/ed25519.cairo
edisontim: Is it alright to have dependencies between different modules of the Alexandria library? I put that there to avoid that

gaetbout: If those dependency become too much of a mess, we might consider to move the most used in a common folder
But I'd try and avoid them as much as possible

edisontim: In my opinion, here it makes sense to use the ArrayTraitExt from data_structures then, there's virtually no risk of a cyclic dependency between ed25519 and array_ext

gaetbout: At this point I'm fine having it but it is good to keep it in mind and monitor those once in a while 🙂

gaetbout:
Where are we at with this PR @edisontim, @LucasLvy ?

LucasLvy:
there are conflicts

edisontim:
> there are conflicts

My bad, solved

LucasLvy (approved):

Issue #138, opened 2023-07-07
feat: Ability to convert a List into an Array handled by the List trait


Bal7hazar:
# Feature Request

**Describe the Feature Request**

List trait is not operable with function that accept Array trait.
To work around this, each user needs to implement his own conversion function.
The request is about to implement this feature directly into the List trait to improve the UX.

**Describe Preferred Solution**

Implement it in the List trait.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

PR #139, opened 2023-07-07
✨ Add List.array() feature

Bal7hazar:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

List trait is not operable with function that accept Array trait.
To work around this, each user needs to implement his own conversion function

Issue Number: #138

## What is the new behavior?

`List.array()` offers a way for users to operate this conversion.

- [ ] Yes
- [x] No

## Other information

Workload ~ 1h

### DIFF
--- a/alexandria/storage/src/list.cairo
+++ b/alexandria/storage/src/list.cairo
@@ -11,7 +11,7 @@ use traits::{Default, DivRem, IndexView, Into, TryInto};
 
 const POW2_8: u32 = 256; // 2^8
 
-#[derive(Drop, Copy)]
+#[derive(Drop)]
 struct List<T> {
     address_domain: u32,
     base: StorageBaseAddress,
@@ -26,7 +26,7 @@ trait ListTrait<T> {
     fn get(self: @List<T>, index: u32) -> Option<T>;
     fn set(ref self: List<T>, index: u32, value: T);
     fn pop_front(ref self: List<T>) -> Option<T>;
-    fn array(ref self: List<T>) -> Array<T>;
+    fn array(self: @List<T>) -> Array<T>;
 }
 
 // when writing elements in storage, we need to know how many storage slots
@@ -119,11 +119,11 @@ impl ListImpl<
         popped
     }
 
-    fn array(ref self: List<T>) -> Array<T> {
+    fn array(self: @List<T>) -> Array<T> {
         let mut array = ArrayTrait::<T>::new();
         let mut index = 0;
         loop {
-            if index == self.len {
+            if index == *self.len {
                 break ();
             }
             array.append(self.get(index).expect('List index out of bounds'));
--- a/alexandria/storage/src/list.cairo
+++ b/alexandria/storage/src/list.cairo
@@ -1,6 +1,7 @@
 use hash::LegacyHash;
 use integer::U32DivRem;
 use option::OptionTrait;
+use array::ArrayTrait;
 use starknet::{storage_read_syscall, storage_write_syscall, SyscallResult, SyscallResultTrait};
 use starknet::storage_access::{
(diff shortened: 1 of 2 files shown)


Bal7hazar:
@milancermak ^

alexandria/storage/src/list.cairo
```diff
@@ -25,6 +26,7 @@ trait ListTrait<T> {
     fn get(self: @List<T>, index: u32) -> Option<T>;
     fn set(ref self: List<T>, index: u32, value: T);
     fn pop_front(ref self: List<T>) -> Option<T>;
+    fn array(ref self: List<T>) -> Array<T>;
```
milancermak, line 29: `array` is a read only op, so this should be `self: @List<T>`

alexandria/storage/src/list.cairo
```diff
@@ -10,7 +11,7 @@ use traits::{Default, DivRem, IndexView, Into, TryInto};
 
 const POW2_8: u32 = 256; // 2^8
 
-#[derive(Drop)]
+#[derive(Drop, Copy)]
```
milancermak, line 14: Is adding `Copy` necessary? What does it do? 🤔

milancermak (changes requested):
Just pointing out `array` should be read only and one question, rest looks good.

alexandria/storage/src/list.cairo
Bal7hazar: I am not very experienced with Copy but I had to add it since I copied value from a ref (self) to an array (reference already moved issue).
Switching to snapshot made it unecessary :+1:

alexandria/storage/src/tests/list_test.cairo
```diff
@@ -14,6 +14,7 @@ trait IAListHolder<TContractState> {
         ref self: TContractState, index: u32, addrs_value: ContractAddress, numbers_value: u256
     );
     fn do_pop_front(ref self: TContractState) -> (Option<ContractAddress>, Option<u256>);
+    fn do_array(ref self: TContractState) -> (Array<ContractAddress>, Array<u256>);
```
milancermak, line 17: This (and related code) should be switched to `self: @TContractState` now as well for it to properly test `array`

Bal7hazar: Woops, my bad

milancermak (approved):
LGTM

alexandria/storage/src/list.cairo
```diff
@@ -116,6 +118,19 @@ impl ListImpl<
 
         popped
     }
+
+    fn array(self: @List<T>) -> Array<T> {
+        let mut array = ArrayTrait::<T>::new();
+        let mut index = 0;
+        loop {
+            if index == *self.len {
+                break ();
```
gaetbout, line 127: There was a bug preventing this (so maybe still there), but you could:
```
break array;
```
 instead? and not returning it at the end of the fn

Bal7hazar: I think I tried, but I got the following issue, I don't think it is worth it, but I could be wrong.
image

gaetbout: y, that is what I thought :/ nvm then :)

LucasLvy: Are you sure it's a bug though ?

gaetbout (approved):

LucasLvy (approved):

alexandria/storage/src/list.cairo
milancermak: I think so. `break`ing from a `loop` has been problematic for a long time, there were multiple bugs already reported and fixed. It works well when returning from outside of the `loop`.

PR #141, opened 2023-07-14
Refactor to use Span

gaetbout:
Passing Array is considered bad practice when not modifying the array.
Moreover it could lead to some ownership issues... It is best to use Span.

Updated some algo to do so

Also took the time to refactor where possible


delaaxe:
agreed as otherwise this requires doing clones.. would be useful to have the `base64` package accepts spans too

src/linalg/dot.cairo
```diff
@@ -24,11 +25,14 @@ fn dot<
...
-        index += 1;
-    };
-    value
+        match xs.pop_front() {
+            Option::Some(x_value) => {
+                let y_value = ys.pop_front().unwrap();
+                value += *x_value * *y_value;
+            },
+            Option::None(_) => {
+                break value;
```
LucasLvy, line 34: why do you return value ?

LucasLvy (changes requested):

gaetbout:
> agreed as otherwise this requires doing clones.. would be useful to have the base64 package accepts spans too

Can't as base 64 needs to add some info sometimes 🙁

src/linalg/dot.cairo
gaetbout: To return the "value" as it was existing before.
Since loop isn't ending by a `;` it'll return that break value as the return of the fn

LucasLvy (approved):

PR #144, opened 2023-07-25
Downgrade to Cairo 2.0.0

delaaxe:
`2.0.1` isn't supported on any testnet so addings alexandria as a dependency to any project makes it undeployable to any network currently.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Build-related changes

### DIFF
--- a/Scarb.toml
+++ b/Scarb.toml
@@ -5,7 +5,7 @@ description = "Starknet utilities"
 homepage = "https://github.com/keep-starknet-strange/alexandria/"
 
 [dependencies]
-starknet = ">=2.0.1"
+starknet = ">=2.0.0"
 
 [scripts]
 all = "scarb build && scarb test"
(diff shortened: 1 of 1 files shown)


delaaxe:
@LucasLvy

LucasLvy:
You can specify a commit/branch using scarb so I don't think it should be on main. I can create a specific branch for that WDYT ?

delaaxe:
Sure, if we can document the syntax for specifying the branch

Scarb.toml
```diff
@@ -5,7 +5,7 @@ description = "Starknet utilities"
 homepage = "https://github.com/keep-starknet-strange/alexandria/"
 
 [dependencies]
-starknet = ">=2.0.1"
+starknet = ">=2.0.0"
```
LucasLvy, line 8: Shouldn't this be like that ?
```suggestion
starknet = "=2.0.0"
```

LucasLvy (changes requested):

LucasLvy:
you probably need to downgrade scarb also

delaaxe:
`2.1` will be on testnet next week so not sure this is needed anymore

PR #145, opened 2023-07-25
Feat/shl use t

gaetbout:
updated shr, shl and fpow to use a trait.
Refactored the code to use into and try_into instead of downcast and upcast (it wasn't possible when this code was written)
Also cut some if branches && removed some irrelevant ref that could increase the amount of steps
This took me about 3h


gaetbout:
@LucasLvy 
I think there is still some work to optimise all 3 of base64 and both sha algorithm but I'd do that in another PR not to add too much at once, and split for each algorithm.
This one is ready

LucasLvy (approved):

PR #154, opened 2023-08-08
feat: migration to workspaces

milancermak:
Introducing Scarb workspaces, separating each module to its own Scarb package.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [x] Build-related changes

Currently, Alexandria is one big package, all modules merged into a single dependency. I think this 1) hurts adoption and 2) stifles development (for example because there's no way how to do breaking changes in a single package). 

## What is the new behavior?

This PR adopts a [workspace](https://docs.swmansion.com/scarb/docs/reference/workspaces) based approach and separates each module to a single package.

- [x] Yes
- [ ] No

Yes in the sense how packages are installed and imported, but no backwards incompatible code changes.

## Other information

There's no material code changes. Most of the PR is just moving files around to a new dir structure. There are some small changes here and there, but mostly are cosmetic (e.g. alphabetical ordering of imports).


milancermak:
cc @mkaput

LucasLvy:
can you rebase por favor

Scarb.toml
```diff
@@ -1,11 +1,18 @@
...
+]
 name = "alexandria"
 version = "0.1.0"
-description = "Starknet utilities"
+description = "Community maintained Cairo and Starknet libraries"
 homepage = "https://github.com/keep-starknet-strange/alexandria/"
 
-[dependencies]
-starknet = ">=2.1.0"
-
```
mkaput, line 9: You can leave it here, and then use `starknet.workspace = true` in particular packages: https://docs.swmansion.com/scarb/docs/reference/workspaces.html#dependencies

src/storage/Scarb.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "alexandria_storage"
+version = "0.1.0"
+description = "Starknet storage utilities"
+homepage = "https://github.com/keep-starknet-strange/alexandria/tree/src/storage"
+
+[[target.starknet-contract]]
+
```
mkaput, line 8:
```suggestion

```

I don't see any contracts here, nor I don't think you'll want to build ones from this package? And not making it `[lib]` makes this package unusable as a dependency: https://docs.swmansion.com/scarb/docs/reference/targets.html#library

mkaput (changes requested):

Scarb.toml
milancermak: Getting this error:

```
error: failed to parse manifest at `/Users/m/projects/my-cairo-forks/alexandria/Scarb.toml`

Caused by:
    this virtual manifest specifies a [dependencies] section, which is not allowed
```

src/storage/Scarb.toml
milancermak: Yes, correct on the contracts part, removing the `[[target.starknet-contract]]` declaration.

Re: `[lib]` - shouldn't this be present in all `Scarb.toml` files in all modules then?

mkaput: > Re: `[lib]` - shouldn't this be present in all `Scarb.toml` files in all modules then?

`[lib]` is the default and is automatically "added" if you don't specify any target :)

Scarb.toml
maciektr: Hi @milancermak!

This needs to be called `[workspace.dependencies]`.
https://docs.swmansion.com/scarb/docs/reference/workspaces.html#dependencies

milancermak: I'm dumb. Thanks.

maciektr: This is just confusing when you do not define the root package, but we mimic this distinction from Cargo manifest format :D

mkaput: I'm making a PR which add a hint how to fix this error FYI :D

milancermak:
@LucasLvy All done amigo

LucasLvy (approved):

PR #156, opened 2023-08-09
feat(storage): v2.1.0 improvements

milancermak:
Additional improvements to `List` implementation.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

## What is the new behavior?

* Removing the custom `StorageSize` trait. It's replaced by `Store::size` since in v2.1.0 it doesn't require an arg anymore.

* Using Poseidon hash instead of Pedersen (LegacyHash) since it's 10x cheaper.

- [x] Yes
- [ ] No

No easy migration path but I'm guessing no one's using it in prod yet. Since it's a breaking change, I did bump the package version to `0.2.0`.

## Other information

Also shortening the `test_append_get_many` test to go over less elements, should be about half as long while still testing the right behavior.

### DIFF
--- a/src/storage/README.md
+++ b/src/storage/README.md
@@ -35,35 +35,12 @@ Note that unlike `get`, using this bracket notation panics when accessing an out
 
 ### Support for custom types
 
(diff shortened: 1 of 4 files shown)


LucasLvy (approved):
lgtm

PR #174, opened 2023-08-31
feat: dedup for arrays

ermvrs:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


src/data_structures/src/array_ext.cairo
```diff
@@ -21,6 +21,7 @@ trait ArrayTraitExt<T> {
     fn index_of_max<impl TPartialEq: PartialEq<T>, impl TPartialOrd: PartialOrd<T>>(
         self: @Array<T>
     ) -> Option<usize>;
+    fn dedup<impl TPartialEq: PartialEq<T>>(self: Array<T>) -> Array<T>;
```
gaetbout, line 24:
```suggestion
    fn dedup<impl TPartialEq: PartialEq<T>>(self: @Array<T>) -> Array<T>;
```

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,43 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+        }
+
+        let mut ret = array![];
+
+        let mut last_value = match self.pop_front() {
+            Option::Some(v) => v,
+            Option::None => {
+                return ret;
+            },
+        };
```
gaetbout, line 154: as you already made sure len >1 you can just pop_front().unwrap()

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,43 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+
+        loop {
+            match self.pop_front() {
+                Option::Some(v) => {
+
+                    if(last_value == v) {
+                        continue;
+                    };
+                    last_value = v;
+                    ret.append(v);
```
gaetbout, line 166: maybe just me but I'd just do:
```suggestion
                    if(last_value != v) {
                      last_value = v;
                      ret.append(v);
                    };
```

src/data_structures/src/tests/array_ext_test.cairo
```diff
@@ -3,6 +3,111 @@ use array::{ArrayTrait, SpanTrait};
 
 use alexandria_data_structures::array_ext::{ArrayTraitExt, SpanTraitExt};
 
+// dedup
```
gaetbout, line 6: Amazing tests :)

src/data_structures/src/array_ext.cairo
ermvrs: Done

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,37 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
     ) -> Option<usize> {
         self.span().index_of_max()
     }
+
+    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: Array<T>) -> Array<T> {
+        let len = self.len();
+
+        if(len <= 1) {
+            return self;
+        }
```
gaetbout, line 145:
```suggestion
        if( self.len() <= 1) {
            return self;
        }
```

gaetbout:
Pinging the chief @LucasLvy 🙏

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,35 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+        let mut ret = array![];
+
+        let mut last_value = self.pop_front().unwrap();
+
+        ret.append(last_value);
+
+        loop {
+            match self.pop_front() {
+                Option::Some(v) => {
+
```
gaetbout, line 154: Maybe get rid of that extra line, but I'm getting picky now 😄

ermvrs: Line: 154 destroyed 💣

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,33 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: Array<T>) -> Array<T> {
+        if(self.len() <= 1) {
+            return self;
+        }
+
+        let mut ret = array![];
+        let mut last_value = self.pop_front().unwrap();
+
+        ret.append(last_value);
```
LucasLvy, line 148:
```suggestion    
        let mut last_value = self.pop_front().unwrap();
        let mut ret = array![last_value];
```

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,33 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+
+        loop {
+            match self.pop_front() {
+                Option::Some(v) => {
+                    if(last_value != v) {
+                      last_value = v;
+                      ret.append(v);
+                    };
+                },
+                Option::None(()) => {
```
LucasLvy, line 158:
```suggestion
                Option::None => {
```

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,33 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
...
+
+        ret.append(last_value);
+
+        loop {
+            match self.pop_front() {
+                Option::Some(v) => {
+                    if(last_value != v) {
+                      last_value = v;
+                      ret.append(v);
+                    };
```
LucasLvy, line 156: This is wrong

LucasLvy (changes requested):
Please:
- Format
- Add a test with `[1, 2, 1]` for example and check that the length is 2.
- The logic seems wrong here you're just checking the next value is not the same as the current one which is not what we would expect from remove duplicates. You'll need to iterate through the array everytime to check if it's already there or not.

gaetbout:
Any reason you reverted the use of @Array ?

ermvrs:
Otherwise i need to clone the array to return at the beginning and use pop_front. Will use more gas :(

ermvrs:
@gaetbout Return type is Array<T> and I also need to use pop_front inside function. So to return will have to same type as parameter in first if case (len < 1). If parameter type is @array then I have to clone it to new array to return that one in first case.
Although further it can be changed if we implement same methods for span as well.

src/data_structures/src/array_ext.cairo
```diff
@@ -135,6 +136,31 @@ impl ArrayImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayTraitExt<T>
     ) -> Option<usize> {
         self.span().index_of_max()
     }
+
+    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: Array<T>) -> Array<T> {
+        if (self.len() <= 1) {
```
gaetbout, line 141:
```suggestion
        if self.len() <= 1 {
```

gaetbout:
You could also do in array_ext
```rust
    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: @Array<T>) -> Array<T> {
        self.span().dedup()
    }
```
And add this to span_ext
```rust
    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: Span<T>) -> Array<T> {
        if self.len() == 0 {
            return array![];
        }
        if (self.len() <= 1) {
            return array![*self[0]];
        }

        let mut last_value = self.pop_front().unwrap();
        let mut ret = array![*last_value];

        loop {
            match self.pop_front() {
                Option::Some(v) => {
                    if (last_value != v) {
                        last_value = v;
                        ret.append(*v);
                    };
                },
                Option::None => {
                    break;
                }
            };
        };

        ret
    }
```

ermvrs:
Done, changed like your suggestion

src/data_structures/src/array_ext.cairo
```diff
@@ -354,4 +360,32 @@ impl SpanImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of SpanTraitExt<T> {
...
         }
     }
+
+    fn dedup<impl TPartialEq: PartialEq<T>>(mut self: Span<T>) -> Array<T> {
+        if self.len() == 0 {
+            return array![];
+        }
+        if (self.len() <= 1) {
+            return array![*self[0]];
+        }
```
gaetbout, line 370: Just realized that this is useless, think you can just removed that

src/data_structures/src/array_ext.cairo
```diff
@@ -43,6 +44,7 @@ trait SpanTraitExt<T> {
     fn index_of_max<impl TPartialEq: PartialEq<T>, impl TPartialOrd: PartialOrd<T>>(
         self: Span<T>
     ) -> Option<usize>;
+    fn dedup<impl TPartialEq: PartialEq<T>>(self: Span<T>) -> Array<T>;
```
LucasLvy, line 47: can you add documentation on what the function does plz

LucasLvy (changes requested):

src/data_structures/src/array_ext.cairo
ermvrs: Where should i add that

LucasLvy: right above the function, rust style doc

LucasLvy (approved):

PR #175, opened 2023-09-01
feat(math): binary exp

milancermak:
Updating the `pow` impl to use binary exponentiation, which is cheaper. 

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

`pow` is O(n)

## What is the new behavior?

`pow` is O(log n)

- [ ] Yes
- [x] No

## Other information

I did a simple [benchmark](https://gist.github.com/milancermak/adf6c70cdd550155a754e335d92fcbf5). Stepwise performance is the same for a simple case, the greater the exponent the more efficient this new implementation is.

Added a testcase to check for the whole 2^N range.

FWIW, it's pretty much the same as `BitShift::fpow` 😁 but without the use of Bitwise ops.


LucasLvy (commented):
I think that we want to keep slow pow for small numbers

milancermak:
> slow pow for small numbers

What do you mean?

LucasLvy:
@enitrat Were you the one saying that the slow power is more efficient than fast power ?

enitrat:
yes but I didn't run the benchmarks yet. Basically the overhead cost brought by additional div/rem opérations outweigh the gains for small exponents

milancermak:
Yeah, if the test runner can be trusted, the small exponents gas usage stays the same.

LucasLvy:
ok you win let's merge it

LucasLvy (approved):

Issue #170, opened 2023-08-31
feat: Merkle tree generation


Bal7hazar:
# Feature Request

**Describe the Feature Request**

The current implementation allows to verify a membership of a merkletree.
It would be interesting to also have the ability to create the whole merkle tree (and prooves) in Cairo.

**Describe Preferred Solution**

[Here](https://github.com/carbonable-labs/whitestark/blob/main/src/merkletree/merkletree.ts) is an implementation I made in JS that we can implement in Cairo.

**Describe Alternatives**

Make it from scratch, but none interest.

**Related Code**

No related code.

**Additional Context**

No additional context.

**If the feature request is approved, would you be willing to submit a PR?**

- [x] Yes
- [ ] No

Issue #169, opened 2023-08-31
feat: Merkle tree based on poseidon


Bal7hazar:
# Feature Request

**Describe the Feature Request**

The current MerkleTree implementation is based on pedersen which is less gas efficient on Starknet regarding poseidon.
The purpose of the feature request is to have a posedion version of the Merkle tree implementation.

**Describe Preferred Solution**

Probably not the final design but something like a generic implementation taking the hash method in parameter.
Then building higher level implementation for each known hash methods using the generic one under the hood.

**Describe Alternatives**

Clone the current implementation and change the hash method.

**Related Code**

[Merkle tree](https://github.com/keep-starknet-strange/alexandria/blob/main/src/data_structures/src/merkle_tree.cairo)

**Additional Context**

No additional context

**If the feature request is approved, would you be willing to submit a PR?**

- [x] Yes
- [ ] No

PR #178, opened 2023-09-08
✨ Merkle tree generation

Bal7hazar:
- [x] Add poseidon version of the merkle tree
- [x] Add proof computation
- [x] Clean and add corresponding tests

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: #169  
Issue Number: #170 

## What is the new behavior?

- A new implementation of the MerkleTree using poseidon hash method
- Add a `compute_proof` method to the MerkleTreeTrait to generate a proof based on leaves and an index
- Rename the previous merkle tree type from MerkleTree to MerkleTreeLegacy
- Merkle tree tests refactoring to clarify tests and add tests according to the new feature

- [x] Yes
- [ ] No

Before the change you could just have:

```cairo
use alexandria_data_structures::merkle_tree::MerkleTreeTrait;
let mut merkle_tree = MerkleTreeTrait::new();
```

Now you must specify which kind of merkle tree you want (the legacy version or the new one):

```cairo
// The new version, using the poseidon hash method
use alexandria_data_structures::merkle_tree::{MerkleTree, HashMethod, MerkleTreeTrait};
let hash_method = HashMethod::Poseidon(());
let mut merkle_tree = MerkleTreeTrait::new(hash_method);
```

```cairo
// The legacy version, using the pedersen hash method
use alexandria_data_structures::merkle_tree::{MerkleTree, HashMethod, MerkleTreeTrait};
let hash_method = HashMethod::Poseidon(());
let mut merkle_tree = MerkleTreeTrait::new(hash_method);
```

## Other information

Work load: ~ 8 hours

### DIFF
--- a/src/data_structures/src/merkle_tree.cairo
+++ b/src/data_structures/src/merkle_tree.cairo
@@ -5,7 +5,7 @@
 //! use alexandria::data_structures::merkle_tree::{MerkleTree, HashMethod, MerkleTreeTrait};
 //!
 //! // Create a new merkle tree legacy instance, this version uses the pedersen hash method.
-//! let hash_method = HashMethod::Pedersen(());
+//! let hash_method = HashMethod::Pedersen;
 //! let mut merkle_tree: MerkleTree = MerkleTreeTrait::new(hash_method);
 //! let mut proof = array![];
 //! proof.append(element_1);
@@ -14,7 +14,7 @@
 //! let root = merkle_tree.compute_root(leaf, proof);
 //!
 //! // Create a new merkle tree instance, this version uses the poseidon hash method.
-//! let hash_method = HashMethod::Poseidon(());
+//! let hash_method = HashMethod::Poseidon;
 //! let mut merkle_tree: MerkleTree = MerkleTreeTrait::new(hash_method);
 //! let mut proof = array![];
 //! proof.append(element_1);
@@ -37,8 +37,8 @@ struct MerkleTree {
 
 #[derive(Serde, Copy, Drop)]
 enum HashMethod {
-    Pedersen: (),
-    Poseidon: (),
+    Pedersen,
+    Poseidon,
 }
 
 /// MerkleTree trait.
@@ -77,7 +77,7 @@ impl MerkleTreeImpl of MerkleTreeTrait<MerkleTree> {
                     // We need to check if the current node is smaller than the current element of the proof.
                     // If it is, we need to swap the order of the hash.
                     current_node = match self.hash_method {
-                        HashMethod::Pedersen(()) => {
+                        HashMethod::Pedersen => {
                             if Into::<felt252, u256>::into(current_node) < (*proof_element).into() {
                                 let mut state = PedersenTrait::new(current_node);
                                 state = state.update(*proof_element);
@@ -88,7 +88,7 @@ impl MerkleTreeImpl of MerkleTreeTrait<MerkleTree> {
                                 state.finalize()
                             }
                         },
-                        HashMethod::Poseidon(()) => {
+                        HashMethod::Poseidon => {
                             let mut state = PoseidonTrait::new();
                             if Into::<felt252, u256>::into(current_node) < (*proof_element).into() {
                                 state = state.update(current_node);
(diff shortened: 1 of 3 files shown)


src/data_structures/src/merkle_tree.cairo
```diff
@@ -54,11 +76,30 @@ impl MerkleTreeImpl of MerkleTreeTrait {
                     // Compute the hash of the current node and the current element of the proof.
                     // We need to check if the current node is smaller than the current element of the proof.
                     // If it is, we need to swap the order of the hash.
-                    if Into::<felt252, u256>::into(current_node) < (*proof_element).into() {
-                        current_node = LegacyHash::hash(current_node, *proof_element);
-                    } else {
-                        current_node = LegacyHash::hash(*proof_element, current_node);
-                    }
+                    current_node = match self.hash_method {
+                        HashMethod::Pedersen(()) => {
```
gaetbout, line 80: Finds this kinda odd to have a match on the the HashMethod, because then if a new hashing gets added it'll break stuff

What is the issue with making it a shared var?

Bal7hazar: Do you suggest to have the state in the struct and do a `self.state.sate = 0` before each hash computation?

Bal7hazar: I think there will be a type issue, since HashState is not a generic Type, I cannot take it as it is, and I'll have to make a struct for each hash method.
Also I don't see why it would break something if a new hash method is added, since it must just add a new enum option and add corresponding matches

Bal7hazar:
@LucasLvy could you take the lead on this review please to move forward?

LucasLvy (changes requested):

src/data_structures/src/merkle_tree.cairo
```diff
@@ -2,63 +2,119 @@
...
+    fn new() -> PoseidonHasher {
+        PoseidonHasher {}
+    }
+    fn hash(ref self: PoseidonHasher, data1: felt252, data2: felt252) -> felt252 {
+        let mut state = PoseidonTrait::new();
+        state = state.update(data1);
+        state = state.update(data2);
+        state.finalize()
+    }
+}
```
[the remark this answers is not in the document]
Bal7hazar: I tried this first and was stuck for long with a `multiple implementation exists` while using `Hasher` in the MerkleTree impl.

src/data_structures/src/merkle_tree.cairo
```diff
@@ -2,63 +2,119 @@
...
-trait MerkleTreeTrait {
+trait MerkleTreeTrait<T> {
     /// Create a new merkle tree instance.
-    fn new() -> MerkleTree;
+    fn new() -> MerkleTree<T>;
     /// Compute the merkle root of a given proof.
-    fn compute_root(ref self: MerkleTree, current_node: felt252, proof: Span<felt252>) -> felt252;
+    fn compute_root(
+        ref self: MerkleTree<T>, current_node: felt252, proof: Span<felt252>
+    ) -> felt252;
```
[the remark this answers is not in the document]
Bal7hazar: What do you mean? Is this because of the generic type? Otherwise this is a the original version: https://github.com/keep-starknet-strange/alexandria/blob/fa3f7e75c3f4074cd0f47cb80c510b5a0916de20/src/data_structures/src/merkle_tree.cairo#L29

src/data_structures/src/merkle_tree.cairo
```diff
@@ -1,64 +1,134 @@
 //! MerkleTree implementation.
 //!
-//! # Example
+//! # Examples
+//!
+//! ```
+//! // This version uses the pedersen hash method because the PedersenHasherImpl is in the scope.
+//! use alexandria_data_structures::merkle_tree::{ Hasher, MerkleTree, pedersen::PedersenHasherImpl, MerkleTreeTrait };
```
LucasLvy, line 7:
```suggestion
//! use alexandria_data_structures::merkle_tree::{Hasher, MerkleTree, pedersen::PedersenHasherImpl, MerkleTreeTrait};
```

src/data_structures/src/merkle_tree.cairo
```diff
@@ -1,64 +1,134 @@
...
+            let mut state = PoseidonTrait::new();
+            state = state.update(data1);
+            state = state.update(data2);
+            state.finalize()
+        }
+    }
+}
 
 /// MerkleTree representation.
+
```
LucasLvy, line 82:
```suggestion
```

LucasLvy (changes requested):

LucasLvy (approved):

PR #179, opened 2023-09-20
feat: keccak256

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

- Implements keccak256 from a bytes input.
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/src/math/src/keccak256.cairo
+++ b/src/math/src/keccak256.cairo
@@ -1,0 +1,65 @@
+use keccak::{cairo_keccak};
+
+#[generate_trait]
+impl U64Impl of U64Trait {
+    /// Converts a little-endian byte slice to a 64-bit unsigned integer
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - A `Span<u8>` slice of size n <=8.
+    ///
+    /// # Returns
+    ///
+    /// A tuple containing the converted 64-bit unsigned integer and the amount of bytes consumed
+    fn from_le_bytes(mut self: Span<u8>) -> (u64, u32) {
+        // Pack full value
+        let mut value: u64 = 0;
+        let mut n_bytes: u32 = self.len();
+        loop {
+            let byte = match self.pop_back() {
+                Option::Some(byte) => *byte,
+                Option::None => {
+                    break;
+                },
+            };
+            value = value * 0x100 + (byte.into());
+        };
+        (value, n_bytes)
+    }
+}
+
+/// Reverse the endianness of an u256
+fn reverse_endianness(value: u256) -> u256 {
+    let new_low = integer::u128_byte_reverse(value.high);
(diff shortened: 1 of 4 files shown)


src/math/src/keccak256.cairo
```diff
@@ -0,0 +1,65 @@
+use keccak::{cairo_keccak};
+
+#[generate_trait]
+impl U64Impl of U64Trait {
+    /// Converts a little-endian byte slice to a 64-bit unsigned integer
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - A `Span<u8>` slice of size n <=8.
```
LucasLvy, line 9: Should this function return a result then if the expected size should be `<= 8`?

src/math/src/keccak256.cairo
```diff
@@ -0,0 +1,65 @@
...
+///
+/// * `self` - A `Array<u8>` of bytes.
+///
+/// # Returns
+///
+/// A `u256` value representing the Keccak hash of the input bytes array.
+fn keccak256(mut self: Array<u8>) -> u256 {
+    // Converts byte array to little endian 8 byte words array.
+    let mut words64: Array<u64> = Default::default();
+    let mut self = self.span();
```
LucasLvy, line 50: Any reason to take an array as argument and convert it straight away to a span ?
```suggestion
fn keccak256(mut self: Span<u8>) -> u256 {
    // Converts byte array to little endian 8 byte words array.
    let mut words64: Array<u64> = Default::default();
```

src/math/src/tests/test_keccak256.cairo
```diff
@@ -0,0 +1,75 @@
...
+fn test_keccak256_empty_bytes() {
+    let input = array![];
+
+    let hash = keccak256(input);
+
+    assert(
+        hash == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470,
+        'wrong hash value'
+    )
+}
```
LucasLvy, line 15: can you really hash nothing ?

LucasLvy (changes requested):

src/math/src/keccak256.cairo
enitrat: nope, thanks

src/math/src/tests/test_keccak256.cairo
enitrat: yes, see https://emn178.github.io/online-tools/keccak_256.html and https://ethereum.stackexchange.com/questions/35103/keccak-hash-of-null-values-result-in-different-hashes-for-different-types

enitrat:
Applied suggestions, added size <=8 assertion

LucasLvy (approved):

Issue #172, opened 2023-08-31
feat: Add method to drop duplicates in ArrayExtTrait


Bal7hazar:
# Feature Request

**Describe the Feature Request**

Droping duplicates would be interesting for many cases, especially after using a concat.

**Describe Preferred Solution**

Loop over the array, `pop_front` elements one by one and use `contains` to identifiy if the element already exists in the shortest array (input or ouput) then add the element in the ouput array.

**Describe Alternatives**

To discuss if needed.

**Related Code**

No related code.

**Additional Context**

Some references discussed in telegram:

- https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique
- https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.dedup

**If the feature request is approved, would you be willing to submit a PR?**

- [x] Yes
- [ ] No

akhercha:
Hey, can I take this one or is it outdated? Thanks!

gaetbout:
@LucasLvy  this can be closed I think

gaetbout:
@LucasLvy This can be closed

PR #181, opened 2023-09-28
feat: unique for arrays

akhercha:
## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Issue number [#172 ](https://github.com/keep-starknet-strange/alexandria/issues/172)

## What is the new behavior?

- Added the `unique` methods for `SpanTraitExt` & `ArrayTraitExt`,
- Added the corresponding unit tests.

- [ ] Yes
- [X] No

## Other information

For this little snippet:
```rust
let shortest: Span<T> = if (self.len() <= 1) || (ret.len() <= self.len()) {
    ret.span()
} else {
    self
};
```

Without the `self.len() <= 1` member some of my tests were failing.
Tried multiple things but couldn't find any better even though I feel like there's a more optimized solution... Curious to have your inputs! 🙏

### DIFF
--- a/src/data_structures/src/array_ext.cairo
+++ b/src/data_structures/src/array_ext.cairo
@@ -395,14 +395,9 @@ impl SpanImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of SpanTraitExt<T> {
     fn unique<impl TPartialEq: PartialEq<T>>(mut self: Span<T>) -> Array<T> {
         let mut ret = array![];
         loop {
-            let shortest: Span<T> = if (self.len() <= 1) || (ret.len() <= self.len()) {
-                ret.span()
-            } else {
-                self
-            };
             match self.pop_front() {
                 Option::Some(v) => {
-                    if !shortest.contains(*v) {
+                    if !ret.contains(*v) {
                         ret.append(*v);
                     }
                 },
--- a/src/data_structures/src/array_ext.cairo
+++ b/src/data_structures/src/array_ext.cairo
@@ -22,7 +22,7 @@ trait ArrayTraitExt<T> {
         self: @Array<T>
     ) -> Option<usize>;
     fn dedup<impl TPartialEq: PartialEq<T>>(self: @Array<T>) -> Array<T>;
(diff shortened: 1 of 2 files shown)


src/data_structures/src/array_ext.cairo
```diff
@@ -45,6 +46,7 @@ trait SpanTraitExt<T> {
         self: Span<T>
     ) -> Option<usize>;
     fn dedup<impl TPartialEq: PartialEq<T>>(self: Span<T>) -> Array<T>;
+    fn drop_duplicates<impl TPartialEq: PartialEq<T>>(self: Span<T>) -> Array<T>;
```
LucasLvy, line 49: Sounds weird to drop duplicates from a span, WDYT @gaetbout @enitrat

gaetbout:
Maybe it is best to keep the rust name (unique) as dev might be more used to that one?

src/data_structures/src/array_ext.cairo
```diff
@@ -385,4 +391,26 @@ impl SpanImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of SpanTraitExt<T> {
...
+                        ret.append(*v);
+                    }
+                },
+                Option::None => {
+                    break;
+                }
+            };
+        };
+        ret
+    }
```
gaetbout, line 415: This pass your tests on my end:
```suggestion
        let mut ret = array![];
        loop {
            match self.pop_front() {
                Option::Some(v) => {
                    if !ret.contains(*v) {
                        ret.append(*v);
                    }
                },
                Option::None => {
                    break;
                }
            };
        };
        ret
    }
```

akhercha: Yea indeed, I had this at first but I was trying to follow the recommendations in the issue:
> Loop over the array, pop_front elements one by one and use contains to identifiy if the element already exists in the shortest array (input or ouput) then add the element in the ouput array.

Thought it would have some wizard optimization magic that I missed.. 😅

LucasLvy (approved):

Issue #158, opened 2023-08-09
feat: storage: `from_array` method for `List`


akirillo:
# Feature Request

**Describe the Feature Request**

I would love to have a simple method, `from_array`, on the `List` interface, which takes in a snapshot of an array and effectively calls `List.append` for every element in it, but only has to write the length to storage once (instead of on every append) since it is known ahead of time.

I've omitted the other feature request template sections since I think the request is pretty straightforward, but if you'd like me to explain in more detail, please let me know

gaetbout:
GM,
Could you point me where is that length being increased multiple times?

PR #183, opened 2023-10-02
feat: from_array for List

akhercha:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: [#158](https://github.com/keep-starknet-strange/alexandria/issues/158)

## What is the new behavior?

- Added the `from_array` method for `ListTrait`
- Added the corresponding unit tests

- [ ] Yes
- [x] No

## Other information

The implementation is very similar to the `append` function except that it sets the length only once.

There's one question still, what should happen when you call the function with an empty list, like `lst.from_array(array![])` ; for my current implementation it "cleans" the list (basically set the length to zero) but maybe this is not what we want. Let me know! 😁

### DIFF
--- a/src/storage/src/list.cairo
+++ b/src/storage/src/list.cairo
@@ -101,24 +101,18 @@ impl ListImpl<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>, impl TStore: Store<T>
     }
 
     fn from_array(ref self: List<T>, array: @Array<T>) {
-        if array.len() == 0 {
-            return;
-        }
         let mut array: Span<T> = array.span();
-        let (base, offset) = calculate_base_and_offset_for_index(
-            self.base, self.len, self.storage_size
-        );
+        let (base, offset) = calculate_base_and_offset_for_index(self.base, 0, self.storage_size);
+        self.len = array.len();
         loop {
             match array.pop_front() {
-                Option::Some(v) => (
-                    Store::write_at_offset(self.address_domain, base, offset, *v).unwrap_syscall()
-                ),
+                Option::Some(v) => (Store::write_at_offset(self.address_domain, base, offset, *v)
+                    .unwrap_syscall()),
                 Option::None => {
                     break;
                 }
             };
         };
-        self.len = array.len();
         Store::write(self.address_domain, self.base, self.len);
     }
 }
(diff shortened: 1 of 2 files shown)


LucasLvy:
`pop` is less efficient than `get` ?

akhercha:
> `pop` is less efficient than `get` ?

With my _limited_ knowledge I'd say `pop` is better but when doing some tests I found that `get` without a `span` had slightly less gaz cosumption than `pop`.
The results (the "test_from_array_big" test contains 44 elements):

image

**First tests** -> using `get`,
**2nd set of tests** -> were using `span` & `pop`.

Or maybe I should not trust the gaz estimation? @LucasLvy

LucasLvy:
interesting then maybe implement, a `from_span` function, and a `from_array` where `from_array` is just:
```rs
fn from_array(arr: Array<T>) { from_span(arr.span()) }
```
(not the exact code but you get the idea)

akhercha:
Updated @LucasLvy !

akhercha:
I also feel like the function can be optimized further by computing `base` & `offset` at the beginning & only re-call `calculate_base_and_offset_for_index` when we overflow the current segment.

Should be something like that:

```rust
fn from_span(ref self: List<T>, mut span: Span<T>) {
        let (mut base, mut offset) = calculate_base_and_offset_for_index(
            self.base, 0, self.storage_size
        );
        self.len = span.len();
        loop {
            match span.pop_front() {
                Option::Some(v) => {
                    Store::write_at_offset(self.address_domain, base, offset, *v).unwrap_syscall();
                    // +1 needed?
                    offset += self.storage_size + 1;
                    // If the current segment overflows, re-compute base & offset
                    if (offset + self.storage_size).into() >= POW2_8 {
                        // then
                        // base is recomputed
                        // & offset becomes 0 
                    }
                },
                Option::None => {
                    break;
                }
            };
        };
        Store::write(self.address_domain, self.base, self.len);
    }
```

But my current attempts made the gas consumption skyrocket 😁

What do you think @LucasLvy ? Shall I continue further or is it not worth it?

LucasLvy:
> Are those gas gains worth in comparison to the loss of clarity in the function? 😅
I'd say no as i'm not even sure how this gas improvement would translate with the step/builtins fees (currently used on the network)

akhercha:
Aaah makes sense. I'll just revert those changes then & we'll see if that's needed in the future. Thanks @LucasLvy !

Got lost in the premature optimizations loop yet again 😎

LucasLvy:
Ok so after asking the sierra gas metering is higher but it consumes less steps/built ins so let’s keep it with pop it’s more optimized for now

PR #199, opened 2023-10-24
Feat: ByteReader and ByteAppender refactored to support generics

sveamarcus:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Currently `.reader()` and `.append_...` only supports `ByteArray`.

Issue Number: N/A

## What is the new behavior?

- Adds a generic implementation divided into two traits
- `ByteReader` - on top of supporting `ByteArray` it can also read from `Array<u8>` and `Span<u8>` making it more versatile
- `ByteAppender` - Refactored all `.append_`... calls into a new trait, which in turn can be used on top of both `Array<u8>` as well as the currently supported `ByteArray`
- Let's illustrate with an example of reading/writing the Bitcoin block header since it is simple enough using both `Array<u8>` and `ByteArray`
```
use alexandria_data_structures::byte_appender::ByteAppender;
use alexandria_data_structures::byte_reader::ByteReader;

// Append
let mut array: Array<u8> = array![];
let version = 1_i32; // strangely a signed int32
let prev_hash = u256{low: 0, high: 0x123};
let merkle_root = u256{low: 0x0, high: 0x456};
let time = 1698179674_u32; 
let n_bits = 0x30c31b18_u32; 
let nonce = fe9f0864_u32;
array.append_i32_le(version); // little endian
array.append_u256(prev_hash); // big endian
array.append_u256(merkle_root); // big endian
array.append_u32_le(time); // little endian
array.append_u32_le(n_bits); // little endian
array.append_u32_le(nonce); // little endian
// this array can now be used to hash the header using sha256

// This is very costly, not sure if it's a good idea to support Into<S, T>? Input welcome. 
// Note, it's not necessary to convert into a ByteArray, .reader() can also be called on
// Array<u8> and Span<u8>. This here is just to show the versatility of the API.
let input: ByteArray = array.into();

// Read a ByteArray containing the header such as above
let mut reader = input.reader();
let version = reader.read_i32_le().unwrap();
let prev_hash = reader.read_u256().unwrap();
let merkle_root = reader.read_u256().unwrap();
let time = reader.read_u32_le().unwrap();
let n_bits = read.read_u32_le().unwrap();
let nonce = reader.read_u32_le().unwrap();
```

- [x] Yes
- [ ] No

Some `use` imports may need updating. All calls remain largely the same.


src/data_structures/src/tests/byte_appender_test.cairo
```diff
@@ -0,0 +1,628 @@
+use alexandria_data_structures::byte_appender::ByteAppender;
+use alexandria_data_structures::tests::byte_reader_test::{
+    test_array_8, test_array_16, test_array_16_neg, test_array_32, test_array_64, test_byte_array_8,
+    test_byte_array_16, test_byte_array_16_neg, test_byte_array_32, test_byte_array_64
+};
+use integer::u512;
+
+
+use debug::PrintTrait;
```
gaetbout, line 9: Woopsie, that should be removed :)

src/data_structures/src/byte_appender.cairo
```diff
@@ -0,0 +1,280 @@
...
+            }
+            let next = (low / one_shift_left_bytes_u128(index - 1)) % 0x100;
+            self.append(next.try_into().unwrap());
+            index -= 1;
+        };
+    }
+
+    fn append_bytes_le(ref self: Array<u8>, bytes: felt252, count: usize) {
+        assert(count <= 16, 'count too big');
+        assert(count <= 16, 'count too big');
```
gaetbout, line 28: duplicate lines

sveamarcus: Thanks for noticing, I pushed an update

PR #209, opened 2023-11-05
refactor: List

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [x] Refactor with API Changes 

Issue Number: N/A

## What is the new behavior?

Refactored the List type to:
- Return SyscallResults instead of panicking in case of storage interaction errors

Added features to make it easier to interact with low-level storage:
- ListTrait::new() to instanciate a new list at a given storage address
- ListTrait::fetch() to fetch a list stored at a given storage address
- Tested these features

Added documentation on List functions
Optimized some functions with inlining

- [x] Yes
- [ ] No

### DIFF
--- a/src/storage/src/list.cairo
+++ b/src/storage/src/list.cairo
@@ -17,119 +17,304 @@ struct List<T> {
 }
 
 trait ListTrait<T> {
+    /// Instantiates a new List with the given base address.
+    ///
+    ///
+    /// # Arguments
+    ///
+    /// * `address_domain` - The domain of the address. Only address_domain 0 is
+    /// currently supported, in the future it will enable access to address
+    /// spaces with different data availability
+    /// * `base` - The base address of the List. This corresponds to the
+    /// location in storage of the List's first element.
+    ///
+    /// # Returns
+    ///
+    /// A new List.
+    fn new(address_domain: u32, base: StorageBaseAddress) -> List<T>;
+
+    /// Fetches an existing List stored at the given base address.
+    /// Returns an error if the storage read fails.
+    ///
+    /// # Arguments
+    ///
+    /// * `address_domain` - The domain of the address. Only address_domain 0 is
+    /// currently supported, in the future it will enable access to address
+    /// spaces with different data availability
+    /// * `base` - The base address of the List. This corresponds to the
+    /// location in storage of the List's first element.
+    ///
+    /// # Returns
+    ///
+    /// An instance of the List fetched from storage, or an error in
+    /// `SyscallResult`.
+    fn fetch(address_domain: u32, base: StorageBaseAddress) -> SyscallResult<List<T>>;
+
+    /// Creates a List from an existing Array. If the array cannot be converted
+    /// to a list due storage errors, an error is returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `address_domain` - The domain of the address. Only address_domain 0 is
+    /// currently supported, in the future it will enable access to address
+    /// spaces with different data availability
+    /// * `address` - The base address of the List. This corresponds to the
+    /// location in storage of the List's first element.
+    /// * `array` - An Array from which to create the List.
+    ///
+    /// # Returns
+    ///
+    /// A List constructed from the array or an error in `SyscallResult`.
+    fn from_array(
+        address_domain: u32, address: StorageBaseAddress, array: @Array<T>
+    ) -> SyscallResult<List<T>>;
+
+    /// Creates a List from an existing Span. Returns an error if the span
+    /// cannot be converted into a list due to storage errors
+    ///
+    /// # Arguments
+    ///
+    /// * `address_domain` - The domain of the address. Only address_domain 0 is
+    /// currently supported, in the future it will enable access to address
+    /// spaces with different data availability
+    /// * `address` - The base address of the List. This corresponds to the
+    /// location in storage of the List's first element.
+    /// * `span` - A Span from which to create the List.
+    ///
+    /// # Returns
+    ///
+    /// A List constructed from the span or an error in `SyscallResult`.
+    fn from_span(
+        address_domain: u32, address: StorageBaseAddress, span: Span<T>
+    ) -> SyscallResult<List<T>>;
+
+    /// Gets the length of the List.
+    ///
+    /// # Returns
+    ///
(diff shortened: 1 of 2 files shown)


enitrat:
> List on the other hand aims to be an abstraction, where its users don't have to care about the underlying storage. By exposing address_domain and StorageBaseAddress the abstraction leaks.

I agree that this abstraction is good to quickly get started using lists, however
- I've been thinking for a while that some of the ways we interact with it are a bit weird. For example, in order to _create_ a list from an array, we have to first _read_ a List, and then pass it to the `from` function. I think it would make so much more sense to just supply a storage variable to create the list at - and the closest abstraction of this is to just pass `variable.address()` to the `from_x` methods. I agree that we lose type safety features by doing this, as we cannot ensure that the variable supplied is actually expecting a `List` type. I wouldn't specifically mind reverting these changes to keep things simple.

- Regarding `address_domain`, I decided to include it in the API to keep it as generic as possible and reduce future breaking changes. 

> They allow for creating storage variables outside of the designated struct Storage which is an expert use case, on the "syscall level" of things 

Yes, these methods enable new ways of using the `List` struct. I don't think that adding features is a bad thing, even if they're complex - I actually think that it's a good thing to provide a complete API that users can use according to their skill level and needs. 

> List is to array what LegacyMap is to dict. A LegacyMap also doesn't offer new, because it's "already there."

I don't really agree in practice - LegacyMap is not a proper type, while `List` is. A LegacyMap doesn't offer `new` because it's not a Cairo type, and thus it cannot be manipulated - it's just a way of computing storage addresses for simple 
variables.
A List should offer `new`, as you need the `List` type to interact with the storage.

Perhaps the delivery of this PR is not optimal, in which case I'd be glad to hear what you would do differently - but I still believe that the underlying idea is good.

LucasLvy:
I would tend to agree with @milancermak, i don't think exposing the address domain is useful (especially yet). Maybe expose it in a different way so you don't have to handle it but you can. As for the `new` method no strong opinion

src/storage/src/list.cairo
```diff
@@ -17,119 +17,304 @@ struct List<T> {
...
+        list.len = span.len();
+
+        let result: SyscallResult<()> = loop {
+            match span.pop_front() {
+                Option::Some(v) => {
+                    let (base, offset) = calculate_base_and_offset_for_index(
+                        list.base, index, list.storage_size
+                    );
+                    match Store::write_at_offset(list.address_domain, base, offset, *v) {
+                        Result::Ok(()) => {},
```
LucasLvy, line 209:
```suggestion
                        Result::Ok(_) => (),
```

src/storage/src/list.cairo
```diff
@@ -17,119 +17,304 @@ struct List<T> {
...
+                        Result::Err(e) => { break Result::Err(e); }
+                    }
+                    index += 1;
+                },
+                Option::None => {
+                    match Store::write(list.address_domain, list.base, list.len) {
+                        Result::Ok(()) => { break Result::Ok(()); },
+                        Result::Err(e) => { break Result::Err(e); }
+                    }
```
LucasLvy, line 218:
```suggestion
                    break Store::write(list.address_domain, list.base, list.len);
```

src/storage/src/list.cairo
```diff
@@ -17,119 +17,304 @@ struct List<T> {
...
+                }
+            };
+        };
+
+        match result {
+            Result::Ok(()) => Result::Ok(list),
```
LucasLvy, line 224:
```suggestion
            Result::Ok(_) => Result::Ok(list),
```

LucasLvy (changes requested):

milancermak:
I think once the `StorageBaseAddress` is exposed in the public API, `address_domain` has to be as well; as Eni mentioned, to future-proof it.

> A LegacyMap doesn't offer new because it's not a Cairo type, and thus it cannot be manipulated - it's just a way of computing storage addresses for simple variables.

I think of List as exactly the same: "a way of computing storage addresses" 😄 It might *technically* be a type because it's a struct (though it cannot be used outside of a Starknet contract), but it's more just a grouping of functions for doing these computations. 

> I actually think that it's a good thing to provide a complete API that users can use according to their skill level and needs.

Good point, yes. 👍 As an expert use case then, `new` and `fetch` make sense to me. Not quite sure about the `from_` functions. First of, I think just `from_span` is sufficient as it's trivial to create a span from an array. But wouldn't it be better to have an `append_span` fn? One could then do `List::new(...).append_span(...)` and the function itself is more practical for general use.

BTW regardless how this PR lands, don't forget to bump the package version number, as it contains breaking changes.

src/storage/src/list.cairo
```diff
@@ -17,119 +17,301 @@ struct List<T> {
...
+    /// * `address_domain` - The domain of the address. Only address_domain 0 is
+    /// currently supported, in the future it will enable access to address
+    /// spaces with different data availability
+    /// * `base` - The base address of the List. This corresponds to the
+    /// location in storage of the List's first element.
+    ///
+    /// # Returns
+    ///
+    /// A new List.
+    fn new(address_domain: u32, base: StorageBaseAddress) -> List<T>;
```
milancermak, line 34: Was wondering if it wouldn't make sense to have `base: felt252` as I imagine most if not all of the usage will be sth like `List::new(0, selector!("my_var_name"))` - no faffing about with `storage_base_address_from_felt252` needed.

enitrat: I'd rather keep it StorageBaseAddress as you can simply do `storage_var.address()`

milancermak: If you have `storage_var`, why is there a need for `new` then? Wouldn't that imply

```rust
struct Storage {
    storage_var: List<T>
}
```
?

LucasLvy:
scarb fmt ftw

LucasLvy:
Format or i'll have to format you

enitrat:
ser give me some time I will address everything at the same time

src/storage/src/list.cairo
enitrat: Given that the expected type for the Store trait is StorageBaseAddress is would expect the user to provide it as well to these functions. felts shouldn't really be exposed

LucasLvy (approved):

PR #216, opened 2023-11-15
chore: bump version

enitrat:
Bumped the version of the package following breaking changes in the API brought by #209

### DIFF
--- a/Scarb.lock
+++ b/Scarb.lock
@@ -62,4 +62,4 @@ version = "0.1.0"
 
 [[package]]
 name = "alexandria_storage"
-version = "0.2.0"
+version = "0.3.0"
--- a/Scarb.toml
+++ b/Scarb.toml
@@ -13,7 +13,7 @@ members = [
     "src/bytes",
 ]
 name = "alexandria"
-version = "0.1.0"
+version = "0.2.0"
 description = "Community maintained Cairo and Starknet libraries"
 homepage = "https://github.com/keep-starknet-strange/alexandria/"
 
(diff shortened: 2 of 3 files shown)


enitrat:
ah, good to know! Indeed it makes sense. Should the virtual workspace have no version at all?

milancermak:
Hmm, I always thought it has to, but it would make sense not to have it there 🤔 Summoning @mkaput to teach us the way 🙏

mkaput:
Frankly? It's actually entirely up to maintainers 😅 

We're identical to Rust in this arena, so there are three possibilities and I'll use Rust projects as examples:
1. Each package has its own `version` that can be unsynchronized (that's what [Tokio](https://github.com/tokio-rs/tokio) and [Gitoxide](https://github.com/Byron/gitoxide) do).
2. You define `workspace.package.version` and then share this version with all workspace members (that's what's happening in [Cairo](https://github.com/starkware-libs/cairo/blob/main/Cargo.toml#L53) repository).
3. You mix both approaches (that's for example what [Scarb](https://github.com/software-mansion/scarb/blob/main/Cargo.toml#L18) [repo](https://github.com/software-mansion/scarb/blob/main/scarb/Cargo.toml#L3) [itself](https://github.com/software-mansion/scarb/blob/main/scarb-metadata/Cargo.toml#L3) does).

IMO, what path you choose depends on the level of coupling of workspace members, and what release strategy you have. In the case of Alexandria, I think the level of coupling is low enough that @milancermak's suggestion makes sense.

There is one drawback to this desynchronized story though: releases are harder because you can't just for-each all members and publish in single dumb loop. Also, you have to manage a bit more Git tags, etc.

LucasLvy (approved):

PR #229, opened 2023-12-11
feat: add ByteArray to storage

glihm:
## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

StarkWare is working on `ByteArray` support. Currently there is no support for `ByteArray` (aka `string`) in storage of contracts.

## What is the new behavior?

This PR introduces a naive implementation for `ByteArray` support in storage.

- simple value in storage.
- as a value of a `LegacyMap` (induces by the first).
- as a key of a `LegacyMap` by deriving `LegacyHash`.

- [ ] Yes
- [X] No

## Other information

About the `Store` trait, I've a doubt on the `size` which returns `1` currently. In the current design, we only store one felt at the base address. Other addresses are derived by a computation of keys.

I bumped to `2.4.0` to have it compiling, all tests are in green. But if it's preferable to have a PR dedicated to the bump, I can rebase later.


src/storage/src/byte_array.cairo
```diff
@@ -0,0 +1,176 @@
...
+impl StoreByteArray of Store<ByteArray> {
+    fn read(address_domain: u32, base: StorageBaseAddress) -> SyscallResult<ByteArray> {
+        StoreByteArray::read_at_offset(address_domain, base, 0)
+    }
+
+    fn write(address_domain: u32, base: StorageBaseAddress, value: ByteArray) -> SyscallResult<()> {
+        StoreByteArray::write_at_offset(address_domain, base, 0, value)
+    }
+
+    fn read_at_offset(
```
LucasLvy, line 68: lots of unwraps here can you change those with `?` especially that this function returns a result

src/storage/src/byte_array.cairo
```diff
@@ -0,0 +1,176 @@
...
+        if data_len == 0 {
+            return Result::Ok(ByteArray { data: array![], pending_word, pending_word_len, });
+        }
+
+        let mut data: Array<bytes31> = array![];
+        let mut index = 0;
+
+        loop {
+            if index == data_len {
+                break ();
```
LucasLvy, line 99:
```suggestion
                break;
```

src/storage/src/tests/byte_array_test.cairo
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    #[external(v0)]
+    impl ImplByteArrayStorage of super::IByteArrayStorage<ContractState> {
+        fn get_b1(self: @ContractState) -> ByteArray {
+            self.b1.read()
+        }
+
+        fn set_b1(ref self: ContractState, value: ByteArray) {
+            self.b1.write(value);
```
LucasLvy, line 35: If the signature of the `write` function changes we want to change this one also no?
```suggestion
            self.b1.write(value)
```

src/storage/src/tests/byte_array_test.cairo
```diff
@@ -0,0 +1,202 @@
...
+            self.map_key.read(key)
+        }
+
+        fn set_map_key(ref self: ContractState, key: ByteArray, value: felt252) {
+            self.map_key.write(key, value);
+        }
+
+        fn health(self: @ContractState) -> felt252 {
+            0x42
+        }
```
LucasLvy, line 64: ?

src/storage/src/byte_array.cairo
glihm: Fixed in 00e14928a0e72b1a8d58c89422fb7e18515c773f.

src/storage/src/byte_array.cairo
glihm: Fixed in 00e14928a0e72b1a8d58c89422fb7e18515c773f, the one left are in a loop, where Cairo does not support `?`.

src/storage/src/tests/byte_array_test.cairo
glihm: Fixed in 827b8c889a291fbf8d84daa08d4b3b3643b8b418. That's a good point, but we will anyway need to update the return type, correct? Or sorry if I don't get the point well.

LucasLvy: exactly

LucasLvy:
Are we storing 1 array / "cell" or are we storing 1 byte / "cell" ?

glihm:
> Are we storing 1 array / "cell" or are we storing 1 byte / "cell" ?

In this proposed PR, the idea is to not rely on segments (perhaps we should?), and use one slot for each element of the `ByteArray` struct and required data.

```rust
ByteArray {
    data: array![1, 2],
    pending_word: 0,
    pending_word_len: 0,
}
```
This will use 5 slots:
```
slot 0: base address (usually the name of the storage variable) => data.len()
slot 1: pending_word
slot 2: pending_word_len
slot 3: data[0]
slot 4: data[1]
```
And the slots addresses are all derived from the base address when `write`/`read` is called, and derived with poseidon from a known set of keys.

Is this something that can work, or we should rework this approach?

LucasLvy:
I'm just wondering the difference between that and the `List<T>` already done, because here it's just a specific usage of the list where `T = Byte31` ?

glihm:
> I'm just wondering the difference between that and the `List<T>` already done, because here it's just a specific usage of the list where `T = Byte31` ?

In a sense yes, agree on that. But on the other hand, the management of the pending word is then totally manual.
With the current solution, the whole data structure is handled by the trait.

LucasLvy:
Does the pending word make sense in this context ?

LucasLvy:
2nd option!
To get the length of the last word we can add a helper function to get the next power of 2. WDYT ?

glihm:
> 2nd option! To get the length of the last word we can add a helper function to get the next power of 2. WDYT ?

I'm really not sure about the DevX in this situation. Users will want to use "strings" (aka ByteArray) in legacy maps too.

Let me know if there is something I'm missing, but it would be like this right?
```rust
    #[storage]
    struct Storage {
        my_string: List<bytes31>,
        strings: List<List<bytes31>>,
    }
```
Where we also need to reconstruct / deconstruct the `ByteArray` to put it in the list with the helper function. So any read/write require a call to an helper function to format the data.

Where we can simply do the following:
```rust
    #[storage]
    struct Storage {
        my_string: ByteArray,
        strings: List<ByteArray>,
    }
```

If it's a matter of efficiency by using the storage segment, I can adapt.
Otherwise, if you do prefer to keep the `bytes31`, feel free to close the PR as it's not relevant anymore as list is already implemented. :+1:

LucasLvy:
I don't have a strong opinion, just want to understand the trade offs here

LucasLvy:
I see where you're going, ok this pr makes sense (misunderstood it)

LucasLvy:
and yes for your second use case you'll need to make sure that you're not writing over a non empty cell so you probably need a segment / bytearray

glihm:
> you need to make sure that you're not overwriting non zero cells. also i still see some unwrap syscalls in functions that return a result

Sorry I shouldn't have resolved the first change request you've made [here](https://github.com/keep-starknet-strange/alexandria/pull/229#discussion_r1422705036) where I added a comment to mention that `?` operator is not supported in loop (at least in `2.4.0`). :+1:

glihm:
> you need to make sure that you're not overwriting non zero cells. also i still see some unwrap syscalls in functions that return a result

And about not re-writing 0 cell, I'm not sure why we need to do that. We rely on poseidon and huge slot space, no?
Because when for instance a user want to update it's value in the storage with a `write`, we inevitably overwrite a non-0 slot.

LucasLvy:
> instance a user want to update it's value in the storage with a write, we inevitably overwrite a non-0 slot.

Yes ofc i just don't want that the byte array in cell `n-1` overwrites the byte array in cell `n`, i didn't check what storage addresses looked like when computed that way (if they're far enough from themselves). If it's far enough i'll be too expensive to store something that would "spill" over the next one so it should be fine

glihm:
> > instance a user want to update it's value in the storage with a write, we inevitably overwrite a non-0 slot.
> 
> Yes ofc i just don't want that the byte array in cell `n-1` overwrites the byte array in cell `n`, i didn't check what storage addresses looked like when computed that way (if they're far enough from themselves). If it's far enough i'll be too expensive to store something that would "spill" over the next one so it should be fine

Totally agree, that's why I've not use the segments, because in my understanding segments are a region of 256 consecutive slots.
So in the proposed PR, I use poseidon for each field of `ByteArray`, and then poseidon again for each element in `data` to ensure as much entropy as possible by selecting a fresh new slot (base address computed with poseidon each time).

LucasLvy:
Now i think that the size isn't correct anymore and should be the length of the `ByteArray` ?

glihm:
> Now i think that the size isn't correct anymore and should be the length of the `ByteArray` ?

Here I have my doubt yes. Because the `ByteArray` has 3 fields, but `data` is an array. So I in the sense I've understood this, the `size` is what we expect in the segment (because it is bounded by `u8` type). And in the current implementation, we always have only `1` felt of the `ByteArray` (data.len()) that is stored in the segment given by the base address. That's why I've put 1.

And the length of `data` of the `ByteArray` can't be known as the `size` method does not take any instance as input. So yes, it's not clear to me what should be this value, as it's only constant size that should be returned by this function.

glihm:
@LucasLvy did you have more inputs about the conformity in the storage for this solution?

LucasLvy:
lol sorry for that they'll add it in the core lib so i guess we don't need to answer this question

glihm:
> https://github.com/starkware-libs/cairo/pull/4596

That even better! Thank you very much and sorry I didn't see it lately.
Will close. :)

PR #230, opened 2023-12-12
feat: generic byte reader support for sha256 reopen

sveamarcus:
## Pull Request type

See previous PR for details

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/src/math/src/sha256.cairo
+++ b/src/math/src/sha256.cairo
@@ -1,13 +1,301 @@
+use alexandria_data_structures::byte_appender::ByteAppender;
+use alexandria_data_structures::byte_reader::{ByteReader, ByteReaderState};
 use integer::{u32_wrapping_add, BoundedInt};
+use traits::DivRem;
 
+#[inline]
+fn sha256<T, +Drop<T>, +ByteReader<T>, +IndexView<ByteReaderState<T>, usize, @u8>>(
+    data: T
+) -> Array<u8> {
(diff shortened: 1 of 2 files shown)


LucasLvy:
I Don't really like changing that. I'd rather keep a sha impl that works with `Array<u8>` directly

sveamarcus:
Fair enough, although ByteArray will probably also need hashing from time to time.

Should I port the reading enhancements over to the Array implementation? Saves over 15% gas cost.

LucasLvy:
what exactly are the changes ? Note that less gas consumption doesn't necessarily equals to less steps

sveamarcus:
Yes, I am aware of the discrepancy between gas cost and proof steps. In this case there is a probable blowup by the amount of data that is rewritten.

In main function `fn sha256(mut data: Array<u8>) -> Array<u8>`, there is a call to `from_u8Array_to_u32Array(data.span())`. This in turn copies the full data array into a new u32 array which is the sha256 block size. In this PR using `ByteReader` semantics the block data is dynamically read out from source byte data. Even though all data is accessed in both cases the additional cost of first creating an `Array<u32>` adds up to the significant difference.

LucasLvy:
So yes please add your optimisations !

PR #231, opened 2023-12-14
feat: byte appender and byte reader support for list storage structure

sveamarcus:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

- Adds support for writing (appender) arbitrary signed and unsigned integer values onto a `List<u8>` of bytes
- Adds reader support from a list of bytes
- Is this actually useful or will the List need to pack bytes onto a `bytes31` first and then `List<bytes31>`?
- Will contracts support storing `ByteArray` directly?

- [ ] Yes
- [ ] No

### DIFF
--- a/Scarb.lock
+++ b/Scarb.lock
@@ -21,6 +21,7 @@ name = "alexandria_data_structures"
 version = "0.2.0"
 dependencies = [
  "alexandria_encoding",
+ "alexandria_storage",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


LucasLvy:
I don't like creating a dependency to the storage crate (it doesn't make sense imo). Also you should check #229 it'd make more sense to do something here

sveamarcus:
Yes, #229 makes more sense. I will take a look and see what I can contribute, if anything.

sveamarcus:
Closing in favor of #229

PR #237, opened 2024-01-05
Base64 Optimization

neotheprogramist:
- optimization of previous base64 encoding algorithm to use less gas
- direct base64 encoding from felt252

### DIFF
--- a/src/encoding/src/base64.cairo
+++ b/src/encoding/src/base64.cairo
@@ -1,4 +1,5 @@
 use alexandria_math::BitShift;
+use alexandria_data_structures::array_ext::ArrayTraitExt;
 use integer::BoundedInt;
 
 const U6_MAX: u128 = 0x3F;
@@ -16,7 +17,7 @@ impl Base64Encoder of Encoder<Array<u8>> {
         let mut char_set = get_base64_char_set();
         char_set.append('+');
         char_set.append('/');
-        inner_encode(data, char_set)
+        encode_u8_array(data, char_set)
     }
 }
 
@@ -25,11 +26,29 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
         let mut char_set = get_base64_char_set();
         char_set.append('-');
         char_set.append('_');
-        inner_encode(data, char_set)
+        encode_u8_array(data, char_set)
     }
 }
 
-fn inner_encode(mut bytes: Array<u8>, base64_chars: Array<u8>) -> Array<u8> {
+impl Base64FeltEncoder of Encoder<felt252> {
+    fn encode(data: felt252) -> Array<u8> {
+        let mut char_set = get_base64_char_set();
+        char_set.append('+');
+        char_set.append('/');
+        encode_felt(data, char_set)
+    }
+}
+
+impl Base64UrlFeltEncoder of Encoder<felt252> {
+    fn encode(data: felt252) -> Array<u8> {
+        let mut char_set = get_base64_char_set();
+        char_set.append('-');
+        char_set.append('_');
+        encode_felt(data, char_set)
+    }
+}
+
+fn encode_u8_array(mut bytes: Array<u8>, base64_chars: Array<u8>) -> Array<u8> {
     let mut result = array![];
     if bytes.len() == 0 {
         return result;
@@ -78,6 +97,52 @@ fn inner_encode(mut bytes: Array<u8>, base64_chars: Array<u8>) -> Array<u8> {
         }
         i += 3;
     };
+    result
+}
+
+fn encode_felt(self: felt252, base64_chars: Array<u8>) -> Array<u8>{
+    let mut result = array![];
+
+    let mut num: u256 = self.into();
+    if num != 0 {
+        let (quotient, remainder) = DivRem::div_rem(
+            num, 65536_u256.try_into().expect('Division by 0')
+        );
+        let remainder: usize = remainder.try_into().unwrap();
+        let r3: usize = (remainder / 1024) & 63;
+        let r2: usize = (remainder / 16) & 63;
+        let r1: usize = (remainder * 4) & 63;
+        result.append(*base64_chars[r1]);
+        result.append(*base64_chars[r2]);
+        result.append(*base64_chars[r3]);
+        num = quotient;
+    }
+    loop {
+        if num == 0 {
+            break;
+        }
+        let (quotient, remainder) = DivRem::div_rem(
+            num, 16777216_u256.try_into().expect('Division by 0')
+        );
+        let remainder: usize = remainder.try_into().unwrap();
+        let r4: usize = remainder / 262144;
+        let r3: usize = (remainder / 4096) & 63;
+        let r2: usize = (remainder / 64) & 63;
+        let r1: usize = remainder & 63;
+        result.append(*base64_chars[r1]);
+        result.append(*base64_chars[r2]);
+        result.append(*base64_chars[r3]);
+        result.append(*base64_chars[r4]);
+        num = quotient;
+    };
+    loop {
+        if result.len() >= 43 {
+            break;
+        }
+        result.append('A');
+    };
+    result = result.reverse();
+    result.append('=');
     result
 }
 
--- a/src/encoding/src/base64.cairo
+++ b/src/encoding/src/base64.cairo
@@ -29,6 +29,58 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
     }
 }
 
+fn inner_encode(mut bytes: Array<u8>, base64_chars: Array<u8>) -> Array<u8> {
+    let mut result = array![];
+    if bytes.len() == 0 {
+        return result;
+    }
+    let mut p: u8 = 0;
+    let c = bytes.len() % 3;
+    if c == 1 {
+        p = 2;
+        bytes.append(0_u8);
+        bytes.append(0_u8);
+    } else if c == 2 {
+        p = 1;
+        bytes.append(0_u8);
+    }
+
+    let mut i = 0;
+    let bytes_len = bytes.len();
+    let last_iteration = bytes_len - 3;
+    loop {
+        if i == bytes_len {
+            break;
+        }
+        let n: u32 = (*bytes[i]).into() * 65536_u32
+            | (*bytes[i + 1]).into() * 256_u32
+            | (*bytes[i + 2]).into();
+        let e1: usize = ((n / 262144) & 63).try_into().unwrap();
+        let e2: usize = ((n / 4096) & 63).try_into().unwrap();
+        let e3: usize = ((n / 64) & 63).try_into().unwrap();
+        let e4: usize = (n & 63).try_into().unwrap();
+        result.append(*base64_chars[e1]);
+        result.append(*base64_chars[(e2)]);
+        if i == last_iteration {
+            if p == 2 {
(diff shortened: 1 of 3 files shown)


LucasLvy:
can you add numbers here plz. At least gas and ideally steps/builtins

ThePinion:
@LucasLvy 
1. The change to the encoding of Array<u8>. Testing on an array of 32 bytes (most use-cases) the old implementation uses 7317558 gas while the new one uses only 804588. It is more than 9 times less. Both implementations have linear complexity so changing the length of the data doesn't affect the relative improvement as tested.
2. Encoding of felt252. The use-case that led to the development of this part was that we needed to encode the transaction hash to base64. Directly encoding the hash (felt252) instead of writing it as an Array<u8> of length 32 in big endian and then encoding leads to a 2.2x improvement (or 11.6x when using the old array encoding function).

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        if i == bytes_len {
+            break;
+        }
+        let n: u32 = (*bytes[i]).into()
+            * 65536_u32 | (*bytes[i + 1]).into()
+            * 256_u32 | (*bytes[i + 2]).into();
+        let e1: usize = ((n / 262144) & 63).try_into().unwrap();
+        let e2: usize = ((n / 4096) & 63).try_into().unwrap();
+        let e3: usize = ((n / 64) & 63).try_into().unwrap();
+        let e4: usize = (n & 63).try_into().unwrap();
```
LucasLvy, line 80:
```suggestion
        let n: usize = (*bytes[i]).into()
            * 65536 | (*bytes[i + 1]).into()
            * 256 | (*bytes[i + 2]).into();
        let e1 = (n / 262144) & 63;
        let e2 = (n / 4096) & 63;
        let e3 = (n / 64) & 63;
        let e4 = n & 63;
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+impl Base64UrlFeltEncoder of Encoder<felt252> {
+    fn encode(data: felt252) -> Array<u8> {
+        let mut char_set = get_base64_char_set();
+        char_set.append('-');
+        char_set.append('_');
+        encode_felt(data, char_set)
+    }
+}
+
+fn encode_u8_array(mut bytes: Array<u8>, base64_chars: Array<u8>) -> Array<u8> {
```
LucasLvy, line 51: Suggest to move to a span as we're not modifying this ?
```suggestion
fn encode_u8_array(mut bytes: Array<u8>, base64_chars: Span<u8>) -> Array<u8> {
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        } else {
+            result.append(*base64_chars[e3]);
+            result.append(*base64_chars[e4]);
+        }
+        i += 3;
+    };
+    result
+}
+
+fn encode_felt(self: felt252, base64_chars: Array<u8>) -> Array<u8> {
```
LucasLvy, line 103: Same here
```suggestion
fn encode_felt(self: felt252, base64_chars: Span<u8>) -> Array<u8> {
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+    let mut result = array![];
+
+    let mut num: u256 = self.into();
+    if num != 0 {
+        let (quotient, remainder) = DivRem::div_rem(
+            num, 65536_u256.try_into().expect('Division by 0')
```
LucasLvy, line 109: It's a constant
```suggestion
            num, 65536_u256.try_into().unwrap()
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        );
+        let remainder: usize = remainder.try_into().unwrap();
```
LucasLvy, line 111: this would deserve either an expect to say why it's safe or a comment above. If not safe then remove it we don't want a library function to panic

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        let r3: usize = (remainder / 1024) & 63;
+        let r2: usize = (remainder / 16) & 63;
+        let r1: usize = (remainder * 4) & 63;
```
LucasLvy, line 114:
```suggestion
        let r3 = (remainder / 1024) & 63;
        let r2 = (remainder / 16) & 63;
        let r1 = (remainder * 4) & 63;
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        result.append(*base64_chars[r2]);
+        result.append(*base64_chars[r3]);
+        num = quotient;
+    }
+    loop {
+        if num == 0 {
+            break;
+        }
+        let (quotient, remainder) = DivRem::div_rem(
+            num, 16777216_u256.try_into().expect('Division by 0')
```
LucasLvy, line 125:
```suggestion
            num, 16777216_u256.try_into().unwrap()
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        );
+        let remainder: usize = remainder.try_into().unwrap();
```
LucasLvy, line 127: same as above

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        let r4: usize = remainder / 262144;
+        let r3: usize = (remainder / 4096) & 63;
+        let r2: usize = (remainder / 64) & 63;
+        let r1: usize = remainder & 63;
```
LucasLvy, line 131:
```suggestion
        let r4 = remainder / 262144;
        let r3 = (remainder / 4096) & 63;
        let r2 = (remainder / 64) & 63;
        let r1 = remainder & 63;
```

src/encoding/src/base64.cairo
```diff
@@ -25,10 +26,126 @@ impl Base64UrlEncoder of Encoder<Array<u8>> {
...
+        result.append(*base64_chars[r4]);
+        num = quotient;
+    };
+    loop {
+        if result.len() >= 43 {
+            break;
+        }
+        result.append('A');
+    };
+    result = result.reverse();
```
LucasLvy, line 144: why not construct the array backward to avoid this reverse ?

src/encoding/src/tests/base64_felt_test.cairo
```diff
@@ -0,0 +1,78 @@
...
+fn bytes_be(val: felt252) -> Array<u8> {
+    let mut result = array![];
+
+    let mut num: u256 = val.into();
+    loop {
+        if num == 0 {
+            break;
+        }
+        let (quotient, remainder) = DivRem::div_rem(
+            num, 256_u256.try_into().expect('Division by 0')
```
LucasLvy, line 17:
```suggestion
            num, 256_u256.try_into().unwrap()
```

src/encoding/src/tests/base64_felt_test.cairo
```diff
@@ -0,0 +1,78 @@
...
+        result.append(remainder.try_into().unwrap());
+        num = quotient;
+    };
+    loop {
+        if result.len() >= 32 {
+            break;
+        }
+        result.append(0_u8);
+    };
+    result = result.reverse();
```
LucasLvy, line 28: same why not backward to avoid a reverse ?

LucasLvy (changes requested):
Either say why the unwraps are safe or remove them and switch to results plz

src/encoding/src/base64.cairo
ThePinion: Yup. fixed (removed unnecessary unwraps)

src/encoding/src/base64.cairo
ThePinion: fixed

src/encoding/src/base64.cairo
ThePinion: fixed

src/encoding/src/base64.cairo
ThePinion: fixed

src/encoding/src/base64.cairo
ThePinion: I thought it was obvious but you're right. I've added a comment: 
```
// Safe since 'remainder' is always less than 65536 (2^16), 
// which is within the range of usize (less than 2^32).
```

src/encoding/src/base64.cairo
ThePinion: fixed

src/encoding/src/base64.cairo
ThePinion: fixed

src/encoding/src/base64.cairo
ThePinion: same as above :))

src/encoding/src/tests/base64_felt_test.cairo
ThePinion: fixed

src/encoding/src/tests/base64_felt_test.cairo
ThePinion: I've thought the additional division required to make this work would outweigh any potential benefit. Upon your suggestion I've tested my hypothesis which turned out to be correct. The code below is around 50% slower compared to one with the ```.reverse()```. 
```cairo
const TWO_TO_248: felt252 = 452312848583266388373324160190187140051835877600158453279131187530910662656;

fn bytes_be_rev(val: felt252) -> Array<u8> {
    let mut result = array![];

    let mut num: u256 = val.into();
    let mut divider: u256 = TWO_TO_248.into();
    loop {
        if num == 0 || divider == 0 {
            break;
        }
        let (quotient, remainder) = DivRem::div_rem(num, divider.try_into().unwrap());
        result.append(quotient.try_into().unwrap());
        num = remainder;
        divider = divider / 256_u256;
    };
    loop {
        if result.len() >= 32 {
            break;
        }
        result.append(0_u8);
    };
    result
}
```
There reason for that is in my opinion the additional assignment of ```divider```` which cancels out with the ```.reverse()``` and the aforementioned division which worsens the performance.

Did you have something other in mind or do you see some potential improvements?

src/encoding/src/base64.cairo
ThePinion: The reasoning is the same as in ```bytes_be``` (see below) although I must confess I've not benchmarked this particular function with its reversed version.

src/encoding/src/tests/base64_felt_test.cairo
LucasLvy: no looks good will merge it as is

LucasLvy (approved):

Issue #214, opened 2023-11-13
feat: Wrapping math operations


Hyodar:
# Feature Request

**Describe the Feature Request**

It's actually pretty simple. Currently, wrapping operations require using the type-specific functions, such as `u64_wrapping_add`, which are not only type-specific but also result in code which is more difficult to understand. Plus, there's no direct operation for wrapping multiplications.

Wrapping operations can be pretty important in some implementations, and as such I feel like they need to have a friendlier interface soon. Even though I feel like this would be a feature corelib should have, I'm not sure what are the current priorities, so maybe it would be better to include it here first. Feel free to direct me to [cairo](https://github.com/starkware-libs/cairo) if you think we could push for adding it to corelib directly.

**Describe Preferred Solution**

A Rust-like interface for wrapping operations in integers.

Basically, adding `WrappingAdd`, `WrappingSub` and `WrappingMul` traits (or `WrappingOps`?) which provide, respectively, `wrapping_add`, `wrapping_sub` and `wrapping_mul`. Those will be implemented for all uints.

`Wrapping` as should be avoided so as to not cause any confusion related to Rust `Wrapping`.

**Describe Alternatives**

**Related Code**

Currently (considering a similar interface for `mul`):
```
u64_wrapping_mul(u64_wrapping_sub(u64_wrapping_add(value1, value2), value3), value4);
```

With this:
```
value1.wrapping_add(value2).wrapping_sub(value3).wrapping_mul(value4);
```

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

enitrat:
Great idea! In Kakarot we're going to implement `CheckedOp` that returns Result<T,T> instead - but it's going to be done in a way similar to your idea.

Hyodar:
Hey! Just added a quick PR for this.

PR #241, opened 2024-01-14
feat: Add WrappingMath trait

Hyodar:
## Add WrappingMath trait

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Individual functions for wrapping ops in uints.

Issue Number: #214

## What is the new behavior?

Adds a `WrappingMath` trait on math and implements it for uints. It provides `wrapping_add`, `wrapping_sub` and `wrapping_mul`, removing the need to use type-specific wrapping operations like `u64_wrapping_add`.

I added it directly to `math`, since I wasn't totally sure where would be the best module to keep this. Feel free to move / ask me to move it somewhere else.

(edit: Also adds individual traits for each operation: `WrappingAdd`, `WrappingSub` and `WrappingMul`)

- [ ] Yes
- [x] No

### DIFF
--- a/src/math/src/lib.cairo
+++ b/src/math/src/lib.cairo
@@ -255,7 +255,7 @@ impl U8WrappingMath of WrappingMath<u8> {
 
     #[inline(always)]
     fn wrapping_mul(self: u8, rhs: u8) -> u8 {
-        (u8_wide_mul(self, rhs) % 0x100_u16).try_into().unwrap()
+        (u8_wide_mul(self, rhs) & BoundedInt::<u8>::max().into()).try_into().unwrap()
     }
 }
 
@@ -272,7 +272,7 @@ impl U16WrappingMath of WrappingMath<u16> {
 
     #[inline(always)]
     fn wrapping_mul(self: u16, rhs: u16) -> u16 {
-        (u16_wide_mul(self, rhs) % 0x10000_u32).try_into().unwrap()
+        (u16_wide_mul(self, rhs) & BoundedInt::<u16>::max().into()).try_into().unwrap()
     }
 }
 
@@ -289,7 +289,7 @@ impl U32WrappingMath of WrappingMath<u32> {
 
     #[inline(always)]
     fn wrapping_mul(self: u32, rhs: u32) -> u32 {
-        (u32_wide_mul(self, rhs) % 0x100000000_u64).try_into().unwrap()
+        (u32_wide_mul(self, rhs) & BoundedInt::<u32>::max().into()).try_into().unwrap()
     }
 }
 
@@ -306,7 +306,7 @@ impl U64WrappingMath of WrappingMath<u64> {
 
     #[inline(always)]
     fn wrapping_mul(self: u64, rhs: u64) -> u64 {
-        (u64_wide_mul(self, rhs) % 0x10000000000000000_u128).try_into().unwrap()
+        (u64_wide_mul(self, rhs) & BoundedInt::<u64>::max().into()).try_into().unwrap()
     }
 }
 
--- a/src/math/src/lib.cairo
+++ b/src/math/src/lib.cairo
@@ -242,85 +242,134 @@ trait WrappingMath<T> {
     fn wrapping_mul(self: T, rhs: T) -> T;
 }
 
-impl U8WrappingMath of WrappingMath<u8> {
+impl WrappingMathImpl<T, +WrappingAdd<T>, +WrappingSub<T>, +WrappingMul<T>> of WrappingMath<T> {
+    #[inline(always)]
+    fn wrapping_add(self: T, rhs: T) -> T {
+        WrappingAdd::<T>::wrapping_add(self, rhs)
+    }
+
+    #[inline(always)]
+    fn wrapping_sub(self: T, rhs: T) -> T {
+        WrappingSub::<T>::wrapping_sub(self, rhs)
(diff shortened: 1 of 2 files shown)


src/math/src/lib.cairo
```diff
@@ -233,3 +235,115 @@ impl U256BitRotate of BitRotate<u256> {
...
+    }
+
+    #[inline(always)]
+    fn wrapping_sub(self: u8, rhs: u8) -> u8 {
+        u8_wrapping_sub(self, rhs)
+    }
+
+    #[inline(always)]
+    fn wrapping_mul(self: u8, rhs: u8) -> u8 {
+        (u8_wide_mul(self, rhs) % 0x100_u16).try_into().unwrap()
```
enitrat, line 258: more efficient to use bitwise operations than using remainder of division

```suggestion
        (u8_wide_mul(self, rhs) & BoundedInt::<u8>::max().into()).try_into().unwrap()
```

src/math/src/lib.cairo
```diff
@@ -233,3 +235,115 @@ impl U256BitRotate of BitRotate<u256> {
...
     }
 }
+
+trait WrappingMath<T> {
+    fn wrapping_add(self: T, rhs: T) -> T;
+    fn wrapping_sub(self: T, rhs: T) -> T;
+    fn wrapping_mul(self: T, rhs: T) -> T;
+}
+
+impl U8WrappingMath of WrappingMath<u8> {
```
enitrat, line 245: I think it makes more sense to have one trait WrappingAdd, WrappingSub, WrappingMul so that if someone wants to create a function with  a requirement on T implementing any of these operations, it doesn't necessarily need to implement the others.

So, for example (equivalent for checkedmath):

```
    trait CheckedMath<T> {
        fn checked_add(self: T, rhs: T) -> Option<T>;
        fn checked_sub(self: T, rhs: T) -> Option<T>;
    }

    impl CheckedMathImpl<T, +CheckedAdd<T>, +CheckedSub<T>> of CheckedMath<T> {
        fn checked_add(self: T, rhs: T) -> Option<T> {
            CheckedAdd::<T>::checked_add(self, rhs)
        }

        fn checked_sub(self: T, rhs: T) -> Option<T> {
            CheckedSub::<T>::checked_sub(self, rhs)
        }
    }
 ```
 
 and then:
 ```
  trait CheckedAdd<T> {
        fn checked_add(self: T, rhs: T) -> Option<T>;
    }

    impl U8CheckedAdd of CheckedAdd<u8> {
        fn checked_add(self: u8, rhs: u8) -> Option<u8> {
            u8_checked_add(self, rhs)
        }
    }
```

(same for other traits)

src/math/src/tests/math_test.cairo
```diff
@@ -224,3 +224,241 @@ fn test_rotr_max() {
     assert(BitRotate::rotate_right(0b101_u128, 127) == 0b1010, 'invalid result');
     assert(BitRotate::rotate_right(0b101_u256, 255) == 0b1010, 'invalid result');
 }
+
+#[test]
+fn test_wrapping_math_non_wrapping() {
+    assert(10_u8.wrapping_add(10_u8) == 20_u8, 'invalid result');
```
enitrat, line 230: instead of doing `assert(a == b, 'error message')` you can do `assert_eq!(a,b)` 

see [Assert Macros](https://book.cairo-lang.org/ch10-01-unrecoverable-errors-with-panic.html?highlight=assert_eq#assert-and-assert_eq-macros)

enitrat (changes requested):

src/math/src/lib.cairo
Hyodar: Makes total sense! Not sure why I hadn't noticed that.

Hyodar: Done in [bb538bc](https://github.com/keep-starknet-strange/alexandria/pull/241/commits/bb538bcdb35c72cc8b75e03cde815f2f67ccd4e5).

src/math/src/lib.cairo
Hyodar: Great, totally agree. With this, we can both provide a full `WrappingMath` trait for succintness but also traits for the individual ops.

Hyodar: Done in [6c6eef1](https://github.com/keep-starknet-strange/alexandria/pull/241/commits/6c6eef1c369c45aa5839c6766dbe0bf7cb123338).

src/math/src/tests/math_test.cairo
Hyodar: Sure! I agree with using the macro here, just kept with the usual `assert` to follow the same standard as the other tests. Still, if you don't think it's a problem using it, that's great.

Hyodar: Done in [6c6eef1](https://github.com/keep-starknet-strange/alexandria/pull/241/commits/6c6eef1c369c45aa5839c6766dbe0bf7cb123338).

enitrat: > Still, if you don't think it's a problem using it, that's great.

yeah basically `assert!` and `assert_eq!` are relatively new - not worth it to update existing tests, but new ones should defintely use it as they're so much more practical!

enitrat (approved):
💯 (but I dont actually have approval rights 😄 )

LucasLvy (approved):

PR #244, opened 2024-01-18
Simplify naming in `i257`

delaaxe:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes

## What is the new behavior?

While code like the following is eventually understandable, it isn't as clearly obvious as the proposed one:

Before:

```rust
fn foo(value: i257) {
  let mut x: u256 = 42;
  if value.sign == true {
    x += value.inner;
  }
}
```

After: 

```rust
fn foo(value: i257) {
  let mut x: u256 = 42;
  if value.is_negative {
    x += value.abs;
  }
}
```

- [X] Yes
- [ ] No

### DIFF
--- a/src/math/src/signed_u256.cairo
+++ b/src/math/src/signed_u256.cairo
@@ -5,8 +5,8 @@
 // The is_negative field is true for negative integers, and false for non-negative integers.
 #[derive(Serde, Copy, Drop, Hash)]
 struct i257 {
-    is_negative: bool,
     abs: u256,
+    is_negative: bool,
 }
 
 #[inline(always)]
--- a/src/math/src/signed_u256.cairo
+++ b/src/math/src/signed_u256.cairo
@@ -3,46 +3,30 @@
 // i257 represents a 129-bit integer.
 // The abs field holds the absolute value of the integer.
 // The is_negative field is true for negative integers, and false for non-negative integers.
-#[derive(Copy, Drop)]
+#[derive(Serde, Copy, Drop, Hash)]
 struct i257 {
     is_negative: bool,
     abs: u256,
 }
 
-
-// Checks if the given i257 integer is zero and has the correct sign.
-// # Arguments
-// * `x` - The i257 integer to check.
-// # Panics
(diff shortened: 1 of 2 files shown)


LucasLvy:
@shramee  any opinion ?

shramee:
Yeah, I think `is_negative` that makes it clearer! ❤️

src/math/src/signed_u256.cairo
```diff
@@ -225,36 +124,192 @@ impl i257PartialOrd of PartialOrd<i257> {
...
+    }
+}
+
+// Adds two i257 integers.
+// # Arguments
+// * `a` - The first i257 to add.
+// * `b` - The second i257 to add.
+// # Returns
+// * `i257` - The sum of `a` and `b`.
+fn i257_add(lhs: i257, rhs: i257) -> i257 {
```
LucasLvy, line 182: why did you move all this functions outside of the traits ?

LucasLvy (changes requested):

src/math/src/signed_u256.cairo
delaaxe: to have a cleaner diff with `i128` which is also used in some projects

delaaxe:
If you're ok I'd also do

```rust
struct i257 {
    abs: u256,
    is_negative: bool,
}
```

instead of 

```rust
struct i257 {
    is_negative: bool,
    abs: u256,
}
```

as it would make storage packing a bit more intuitive if ever needed

src/math/src/signed_u256.cairo
delaaxe: or at least, the corelib implementation wasn't sufficient enough and that's why it needed to be redone in satoru iirc

LucasLvy: keep them in the traits plz

delaaxe: done

LucasLvy (approved):

PR #262, opened 2024-01-25
Fix: `i257` negative zeros

delaaxe:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


LucasLvy (changes requested):
I think that for clarity in error messages you shouldn't use the regular `Add` trait of `u256` but rather [this](https://github.com/starkware-libs/cairo/blob/fe0ca27d38b44e60b1d1653c0dbf12199fe80d59/corelib/src/integer.cairo#L1066-L1116) and use `.expect("useful error message")` on the result

delaaxe:
> I think that for clarity in error messages you shouldn't use the regular `Add` trait of `u256` but rather [this](https://github.com/starkware-libs/cairo/blob/fe0ca27d38b44e60b1d1653c0dbf12199fe80d59/corelib/src/integer.cairo#L1066-L1116) and use `.expect("useful error message")` on the result

done, i think you meant those functions

LucasLvy:
No i was referring to the functions i linked (`u256_overflowing_add`,`u256_overflowing_sub` etc)

delaaxe:
those are clumsier to use, the checked ones allow using expect more directly. unless i completely misunderstood what you wanted to do?

LucasLvy:
`checked_add` calls `overflowing_add` so you get a small improvement. It shouldn't be hard to use just do
```rs
let (r, overflow) = u256_overflowing_add(lhs, rhs);
let r = r.expect("....");
```

delaaxe:
the first result value is a `u256`, so there's no `expect` method on it. why reimplement a function that already exists (the checked one)?

delaaxe:
@LucasLvy can you check please?

LucasLvy:
correct leave that as it is

LucasLvy (approved):

delaaxe:
Thanks boss

PR #277, opened 2024-02-19
Fix merkle tree hashing for poseidon

Leonard-Pat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

The current Poseidon hashing implementation for the merkle tree is incorrect. It uses 
```
            let mut state = PoseidonTrait::new();
            state = state.update(data1);
            state = state.update(data2);
            state.finalize()

```
Under the hood this actually hashes the elements twice  (once from the state update and once from finalization, see [corelib](https://github.com/starkware-libs/cairo/blob/main/corelib/src/poseidon.cairo)), the intended behaviours is that only one round of hashing is done (when hashing two elements). This is noted in https://docs.starknet.io/documentation/architecture_and_concepts/Cryptography/hash-functions/#poseidon_hash where by the hash of two elements is simply 
Screenshot 2024-02-19 at 11 53 04. 
Also please see POC https://github.com/Leonard-Pat/merkle-tree-failing 

- Fixes Poseidon hashing for merkle tree poseidon impl

- [ ] Yes
- [x] No

## Other information

Updated tests to reflect this change - note the gas limit was exceeded and therefore needed to be udpated as well

### DIFF
--- a/src/merkle_tree/src/merkle_tree.cairo
+++ b/src/merkle_tree/src/merkle_tree.cairo
@@ -41,7 +41,6 @@ struct Hasher {}
 
 mod pedersen {
     use hash::HashStateTrait;
-    use pedersen::PedersenTrait;
     use super::{Hasher, HasherTrait};
 
     impl PedersenHasherImpl of HasherTrait<Hasher> {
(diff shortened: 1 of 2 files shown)


LucasLvy (changes requested):
can you also fix pedersen plz

Leonard-Pat:
> can you also fix pedersen plz

Pedersen is i believe is correct as the finalization step does not add another round of hashing

LucasLvy:
ah indeed but i think that using directly the `pedersen` function would reduce a bit the step cost

Leonard-Pat:
> ah indeed but i think that using directly the `pedersen` function would reduce a bit the step cost

okay done :)

LucasLvy (approved):

Issue #274, opened 2024-02-08
feat: add: Solidity-like abi.encode functions


b-j-roberts:
# Feature Request

It would be helpful to have utilities to emulate the following functionalities that solidity provides : 
- `abi.encode`
- `abi.encodePacked`

There are some other ABI functions, too, which are less valuable in the immediate term:
- `abi.encodeWithSelector`
- `abi.decode`
- `abi.encodeWithSignature`
- ...

These would be super helpful in the [Blobstream Cairo contracts](https://github.com/keep-starknet-strange/blobstream-starknet/tree/main), which are porting the Blobstream solidity contracts.
[Corresponding issue](https://github.com/keep-starknet-strange/blobstream-starknet/issues/69)

**Preferred Solution**

It's a little tricky because Cairo 1 doesn't support variadic arguments. I think the preferred solution would be something like : 
```
trait EthAbiEncodeTrait<T> {
    fn encode(self: Bytes, x: T) -> Bytes;
    fn encodePacked(self: Bytes, x: T) -> Bytes;
}
```
With the trait implemented for each type. Allowing us to use it like this : 
```
let x32: u32 = 100;
let x64: u64 = 200;
let mut bytes: Bytes = BytesTrait::new_empty();
bytes.encodePacked(x64).encodePacked(x32)
```

[More info here](https://github.com/keep-starknet-strange/blobstream-starknet/issues/69).

**Alternatives**

Currently, there exists the alternative of just using the `BytesTrait` in `alexandria` with the `append_X` functions as described in the [README.md](https://github.com/keep-starknet-strange/alexandria/blob/main/src/bytes/README.md). But, this is more error-prone because users need to understand the nuances of the `encode` function.

**Related Code**

[Example here.](https://github.com/keep-starknet-strange/blobstream-starknet/issues/69#issuecomment-1932558985)

**If the feature request is approved, would you be willing to submit a PR?**

- [x] Yes
- [ ] No

drspacemn:
@LucasLvy thoughts on this addition?

LucasLvy:
sure but add that in a new crate ?

b-j-roberts:
Sure. A couple possible names for the new crate might be `sol_tools`, `sol_utils`, `eth_tools`, `eth_utils`, or maybe someone has a better idea?

I think the existing `encoding` crate could be good too. I could add a module called `abi`, `eth_abi`, or `sol_abi`.

LucasLvy:
[...]

PR #279, opened 2024-02-22
feat: Solidity ABI encode functions and utils

b-j-roberts:
Added various traits/utils for supporting Solidity-like ABI encode functions. This includes the traits :
- `SolAbiEncodeTrait` : providing equivalents for Solidity `abi.encode` and `abi.encodePacked`
- `SolAbiEncodeSelectorTrait` : for Solidity `abi.encodeWithSelector`
- `SolAbiDecodeTrait` : for Solidity `abi.decode`
- `SolBytesTrait` : Providing utilities for easy conversion of types into Solidity-like `bytes1`, `bytes2`, ..., `bytes32` types

Also added some helpers for `alexandria_bytes::Bytes`, such as : 
- `Bytes` <-> Cairo's `ByteArray` & `bytes31` function options
- Debug & Display traits for `Bytes`, to allow printing `Bytes` as a hex string

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: #274 

- [ ] Yes
- [x] No

### DIFF
--- a/src/bytes/src/bytes.cairo
+++ b/src/bytes/src/bytes.cairo
@@ -165,17 +165,18 @@ impl BytesImpl of BytesTrait {
     fn to_byte_array(self: Bytes) -> ByteArray {
         let mut res: ByteArray = Default::default();
         let mut offset = 0;
-        while offset < self.size() {
-            if offset + 31 <= self.size() {
-                let (new_offset, value) = self.read_bytes31(offset);
-                res.append_word(value.into(), 31);
-                offset = new_offset;
-            } else {
-                let (new_offset, value) = self.read_u8(offset);
-                res.append_byte(value);
-                offset = new_offset;
-            }
-        };
+        while offset < self
+            .size() {
+                if offset + 31 <= self.size() {
+                    let (new_offset, value) = self.read_bytes31(offset);
+                    res.append_word(value.into(), 31);
+                    offset = new_offset;
+                } else {
+                    let (new_offset, value) = self.read_u8(offset);
+                    res.append_byte(value);
+                    offset = new_offset;
+                }
+            };
         res
     }
 
--- a/src/bytes/src/bytes.cairo
+++ b/src/bytes/src/bytes.cairo
@@ -1,8 +1,8 @@
 use alexandria_bytes::utils::{
     u128_join, read_sub_u128, u128_split, u128_array_slice, keccak_u128s_be, u8_array_to_u256
 };
-use alexandria_math::{U128BitShift, U256BitShift};
 use alexandria_math::sha256::sha256;
+use alexandria_math::{U128BitShift, U256BitShift};
 use starknet::ContractAddress;
 
 /// Bytes is a dynamic array of u128, where each element contains 16 bytes.
--- a/src/bytes/src/bytes.cairo
+++ b/src/bytes/src/bytes.cairo
@@ -146,9 +146,11 @@ impl BytesImpl of BytesTrait {
 
     fn from_byte_array(mut bytes: ByteArray) -> Bytes {
         let mut res = BytesTrait::new_empty();
-        while !bytes.data.is_empty() {
-            let value: bytes31 = bytes.data.pop_front().unwrap();
-            res.append_bytes31(value);
+        loop {
+            match bytes.data.pop_front() {
+                Option::Some(val) => res.append_bytes31(val),
+                Option::None => { break; }
+            }
         };
         // Last elem
         if bytes.pending_word_len != 0 {
@@ -160,7 +162,7 @@ impl BytesImpl of BytesTrait {
     }
 
     fn to_byte_array(self: Bytes) -> ByteArray {
-        let mut res: ByteArray = "";
+        let mut res: ByteArray = Default::default();
         let mut offset = 0;
         while offset < self.size() {
             if offset + 31 <= self.size() {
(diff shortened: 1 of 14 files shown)


src/bytes/src/bytes.cairo
```diff
@@ -135,6 +144,38 @@ impl BytesImpl of BytesTrait {
         Bytes { size, data }
     }
 
+    fn from_byte_array(mut bytes: ByteArray) -> Bytes {
+        let mut res = BytesTrait::new_empty();
+        while !bytes.data.is_empty() {
+            let value: bytes31 = bytes.data.pop_front().unwrap();
+            res.append_bytes31(value);
+        };
```
LucasLvy, line 152:
```cairo
loop {
    match bytes.data.pop_front() {
        Option::Some(...) => ...,
        Option::None => break,
    }
}
```
is better currently

src/bytes/src/bytes.cairo
```diff
@@ -135,6 +144,38 @@ impl BytesImpl of BytesTrait {
...
+        // Last elem
+        if bytes.pending_word_len != 0 {
+            let mut val: u256 = bytes.pending_word.into();
+            val = U256BitShift::shl(val, 8 * (32 - bytes.pending_word_len.into()));
+            res.concat(@BytesTrait::new(bytes.pending_word_len, array![val.high, val.low]));
```
LucasLvy, line 157: why not just `append_u256` ?

src/bytes/src/bytes.cairo
```diff
@@ -135,6 +144,38 @@ impl BytesImpl of BytesTrait {
...
+        }
+        res
+    }
+
+    fn to_byte_array(self: Bytes) -> ByteArray {
+        let mut res: ByteArray = "";
```
LucasLvy, line 163: If it works then that'd be beeter
```suggestion
        let mut res: ByteArray = Default::default();
```

src/bytes/src/bytes.cairo
```diff
@@ -358,6 +399,19 @@ impl BytesImpl of BytesTrait {
...
 
+    /// read bytes31 from Bytes
+    #[inline(always)]
+    fn read_bytes31(self: @Bytes, offset: usize) -> (usize, bytes31) {
+        let (new_offset, high) = self.read_u128(0);
+        let (new_offset, low) = self.read_u128_packed(new_offset, 15);
+        let low = U128BitShift::shl(low, 8);
+        let mut value: u256 = u256 { high, low };
+        value = U256BitShift::shr(value, 8);
+        let value: felt252 = value.try_into().expect('Couldn\'t convert to felt252');
```
LucasLvy, line 410: add a comment to say why it's safe to unwrap plz

src/bytes/src/utils.cairo
```diff
@@ -1,7 +1,62 @@
...
 use keccak::{u128_to_u64, u128_split as u128_split_to_u64, cairo_keccak};
 
+fn format_byte_hex(byte: u8, ref f: Formatter) -> Result<(), Error> {
+    let base: NonZero<u8> = 16_u8.try_into().unwrap();
+    if byte < 0x10 {
+        let zero: ByteArray = "0";
+        let res = Display::fmt(@zero, ref f);
+        if res.is_err() {
+            return res;
+        }
```
LucasLvy, line 15: not sure i understand that

src/bytes/src/utils.cairo
```diff
@@ -1,7 +1,62 @@
...
+}
+
+impl BytesDebug of Debug<Bytes> {
+    fn fmt(self: @Bytes, ref f: Formatter) -> Result<(), Error> {
+        let mut i: usize = 0;
+        let prefix: ByteArray = "0x";
+        let mut res = Display::fmt(@prefix, ref f);
+        if res.is_err() {
+            return res;
+        }
```
LucasLvy, line 27: same here

src/bytes/src/utils.cairo
```diff
@@ -1,7 +1,62 @@
...
+        while i < self.size() {
+            let (new_i, value) = self.read_u8(i);
+            res = format_byte_hex(value, ref f);
+            if res.is_err() {
+                break;
+            }
```
LucasLvy, line 33: same

LucasLvy (changes requested):

src/bytes/src/bytes.cairo
b-j-roberts: If we append_u256, the size of `Bytes` will increase by 32. So the resulting Bytes object will have trailing 0s. Only appending the `pending_word_len` bytes allows us to use `from_byte_array` and `to_byte_array` as inverses of each other without getting weird 0s added to the end/beginning.

src/bytes/src/utils.cairo
LucasLvy: just use `?`

src/bytes/src/utils.cairo
```diff
@@ -1,7 +1,63 @@
...
+        let prefix: ByteArray = "0x";
+        let mut res = Display::fmt(@prefix, ref f);
+        if res.is_err() {
+            return res;
+        }
+        while i < self.size() {
+            let (new_i, value) = self.read_u8(i);
+            res = format_byte_hex(value, ref f);
+            if res.is_err() {
+                break;
```
LucasLvy, line 33: If res is an error do you really want to return the last correct res or do you also want to propagate the error ?

src/bytes/src/utils.cairo
```diff
@@ -1,7 +1,63 @@
...
```
LucasLvy, line 53: If res is an error do you really want to return the last correct res or do you also want to propagate the error ?

LucasLvy (changes requested):

src/bytes/src/utils.cairo
b-j-roberts: I think propagating the error would be ideal. Otherwise, the user gets incomplete hex representations that cause confusion.
Is this the best way to do that, since you cannot use `?` in a loop?

LucasLvy: ah yes i always forget

LucasLvy: so return the error instead of just breaking ? or `break res` and propagate the error after

PR #282, opened 2024-03-01
Fixing i257 is_zero()

gaetbout:

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Fixing is_zero behaviour that was causing is_zero to sometimes return false when it was indeed zero

### DIFF
--- a/src/math/src/i257.cairo
+++ b/src/math/src/i257.cairo
@@ -259,9 +259,11 @@ impl i257Zeroable of Zeroable<i257> {
         i257_new(0, false)
     }
     fn is_zero(self: i257) -> bool {
+        assert(!self.is_negative, 'no negative zero');
         self.abs == 0
     }
     fn is_non_zero(self: i257) -> bool {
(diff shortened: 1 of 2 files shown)


delaaxe:
this solution preserves negative zeros which shouldn't be allowed

gaetbout:
You saying it shouldn't be possible to create `i257 { abs: 0, is_negative: true };` ?
I'm afraid this would add a lot of `Result` to handle that case then

delaaxe:
Check out the other functions, they already panic when that's the case

LucasLvy (changes requested):
should panic for negative zero

gaetbout:
How should I enforce that?
The `i257_new()` fn already enforces that it can't be done, if the user creates it manually, there is nothing I can do.
Am I missing something?

LucasLvy:
Add and assertion in the function

LucasLvy (approved):

PR #288, opened 2024-03-28
feat: Bytes Store

b-j-roberts:
Add `Store` implementation for the `Bytes` type to allow using `Bytes` in contract storage. Also added test cases.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Currently, when trying to use `Bytes` in contract storage you get errors like this :
```
starknet::Store::<Bytes>::write(
                          ^***^
Error: Failed to compile test artifact, for detailed information go through the logs above
```

Issue Number: N/A

## What is the new behavior?

- Add storage functions to allow using `Bytes` in storage
- `Bytes` stored in the following format :
  - Only the size in bytes is stored in the original address where the `Bytes` object is stored.
  - The actual data is stored in chunks of 256 `u128` values in another location in storage determined by the hash of:
    - The address storing the size of the bytes object.
    - The chunk index.
    - The short string "Bytes"

I used a very similar implementation to the `ByteArray` storage functions from [cairo/corelib](https://github.com/starkware-libs/cairo/blob/d090fec472ee3fc40f04b1a19bc362e7bb18e219/corelib/src/starknet/storage_access.cairo#L677)

- [ ] Yes
- [x] No

## Other information

This would be useful since it is a common pattern in smart contract programming to use storage for accessing values between requests/calls and callbacks.

For example, this would help a lot in implementing `succinct-starknet`, where a `Bytes` output is [passed to a callback through storage](https://github.com/succinctlabs/succinctx/blob/9df6a9db651507d60ffa2d75eda3fe526d13f90a/contracts/src/SuccinctGateway.sol#L272).

### DIFF
--- a/src/bytes/src/storage.cairo
+++ b/src/bytes/src/storage.cairo
@@ -1,0 +1,146 @@
+use alexandria_bytes::bytes::{Bytes, BytesTrait, BYTES_PER_ELEMENT};
+use starknet::SyscallResult;
+use starknet::storage_access::{
+    Store, StorageAddress, StorageBaseAddress, storage_address_from_base,
+    storage_base_address_from_felt252, storage_address_from_base_and_offset
+};
+
+/// Store for a `Bytes` object.
+///
+/// The layout of a `Bytes` object in storage is as follows:
+/// * Only the size in bytes is stored in the original address where the
+///   bytes object is stored.
+/// * The actual data is stored in chunks of 256 `u128` values in another location
+///   in storage determined by the hash of:
+///   - The address storing the size of the bytes object.
+///   - The chunk index.
+///   - The short string `Bytes`.
+pub impl BytesStore of Store<Bytes> {
+    #[inline(always)]
+    fn read(address_domain: u32, base: StorageBaseAddress) -> SyscallResult<Bytes> {
+        inner_read_bytes(address_domain, storage_address_from_base(base))
+    }
+    #[inline(always)]
+    fn write(address_domain: u32, base: StorageBaseAddress, value: Bytes) -> SyscallResult<()> {
+        inner_write_bytes(address_domain, storage_address_from_base(base), value)
+    }
+    #[inline(always)]
+    fn read_at_offset(
+        address_domain: u32, base: StorageBaseAddress, offset: u8
+    ) -> SyscallResult<Bytes> {
+        inner_read_bytes(address_domain, storage_address_from_base_and_offset(base, offset))
+    }
+    #[inline(always)]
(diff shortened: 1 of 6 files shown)


src/bytes/src/storage.cairo
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/// Returns a pointer to the `chunk`'th of the Bytes object at `address`.
+/// The pointer is the `Poseidon` hash of:
+/// * `address` - The address of the Bytes object (where the size is stored).
+/// * `chunk` - The index of the chunk.
+/// * The short string `Bytes` is used as the capacity argument of the sponge
+///   construction (domain separation).
+fn inner_bytes_pointer(address: StorageAddress, chunk: felt252) -> StorageBaseAddress {
+    let (r, _, _) = core::poseidon::hades_permutation(address.into(), chunk, 'Bytes'_felt252);
```
LucasLvy, line 52: This doesn't work ? with the suffix it's pretty confusing
```suggestion
    let (r, _, _) = core::poseidon::hades_permutation(address.into(), chunk, 'Bytes');
```

src/bytes/src/storage.cairo
```diff
@@ -0,0 +1,147 @@
...
+        size, BYTES_PER_ELEMENT.try_into().unwrap()
+    );
+    let mut chunk = 0;
+    let mut chunk_base = inner_bytes_pointer(address, chunk);
+    let mut index_in_chunk = 0_u8;
+    let mut result: Bytes = BytesTrait::new_empty();
+    loop {
+        if remaining_full_words == 0 {
+            break Result::Ok(());
+        }
```
LucasLvy, line 75: this looks like a while loop ?

src/bytes/src/storage.cairo
```diff
@@ -0,0 +1,147 @@
...
+    starknet::syscalls::storage_write_syscall(address_domain, address, size.into())?;
+    let mut words = value.data.span();
+    let mut chunk = 0;
+    let mut chunk_base = inner_bytes_pointer(address, chunk);
+    let mut index_in_chunk = 0_u8;
+    loop {
+        let curr_value = match words.pop_front() {
+            Option::Some(x) => x,
+            Option::None => { break Result::Ok(()); },
+        };
```
LucasLvy, line 126: convert that to a `while let Option::Some(curr_value) = works.pop_front()`

LucasLvy (changes requested):
otherwise looks good

LucasLvy (approved):

PR #297, opened 2024-04-22
Rework sorting

gaetbout:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Rework the sorting package to introduce a common API that all sorting algorithm should follow from now on.
BREAKING CHANGES: API

### DIFF
--- a/src/sorting/src/bubble_sort.cairo
+++ b/src/sorting/src/bubble_sort.cairo
@@ -8,9 +8,12 @@
 use super::Sortable;
 
 pub impl BubbleSort of Sortable {
-    fn sort<T, +Copy<T>, +Drop<T>, +PartialOrd<T>, +PartialEq<T>>(mut array: Array<T>) -> Array<T> {
-        if array.len() <= 1 {
-            return array;
+    fn sort<T, +Copy<T>, +Drop<T>, +PartialOrd<T>>(mut array: Span<T>) -> Array<T> {
+        if array.len() == 0 {
+            return array![];
+        }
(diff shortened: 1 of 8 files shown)


gaetbout:
@LucasLvy ?

src/sorting/src/bubble_sort.cairo
```diff
@@ -5,39 +5,44 @@
...
-            idx2 = 1;
-            sorted_iteration = 0;
-        } else {
-            if (*array[idx1] == *array[idx2]) || !((asc) ^ (*array[idx1] < *array[idx2])) {
+        loop {
+            if idx2 == array.len() {
                 sorted_array.append(*array[idx1]);
-                idx1 = idx2;
-                idx2 += 1;
+                if sorted_iteration {
```
LucasLvy, line 26: isn't this a while ? (do we want to keep the bubble sort ?)

LucasLvy (approved):

src/sorting/src/bubble_sort.cairo
gaetbout: wdym?
Putting this as the while condition to exit?

LucasLvy: yes

gaetbout: can't because it also depends on whether `idx2 == array.len()`.
Or maybe we can?
Didn't try much to update the logic in this case

PR #307, opened 2024-05-17
ed25519: remove extended homogenous point

shramee:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

No changes in the behaviour, just cleaning up old homogenous extended point code.

- [ ] Yes
- [x] No

### DIFF
--- a/src/math/src/ed25519.cairo
+++ b/src/math/src/ed25519.cairo
@@ -65,14 +65,6 @@ pub struct Point {
     pub y: u256,
 }
 
-#[derive(Drop, Copy)]
-pub struct ExtendedHomogeneousPoint {
-    pub X: u256,
-    pub Y: u256,
-    pub Z: u256,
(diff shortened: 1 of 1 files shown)


tdelabro:
@shramee sorry @LucasLvy is in hollyday. 
Can you explain why is this a good idea? Are homogenous point not a thing anymore

shramee:
Projective coordinates are used for optimized curve point operations on traditional CPUs where they assume cost of 1 Inv > 100 muls
For us in Cairo, inv is actually cheaper than mul followed by a div_mod.
image
For this reason, Affine coords perform better than projective coords in hint based finite field architectures.
Savings are quite decent for us too.
So while homogenous points are still a thing, I think we can clear away the dead code we won't use.

tdelabro (approved):

PR #308, opened 2024-05-18
optim sha512 :rocket: :rocket: :rocket:

shramee:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

```diff
< [PASS] alexandria_math::tests::sha512_test::test_size_zero (gas: ~5698)
> [PASS] alexandria_math::tests::sha512_test::test_size_zero (gas: ~632)
<         steps: 735949
>         steps: 85153
<         memory holes: 73172
>         memory holes: 4849
<         builtins: ("range_check_builtin": 142428, "bitwise_builtin": 2025)
>         builtins: ("range_check_builtin": 15797, "bitwise_builtin": 2702)

< [PASS] alexandria_math::tests::sha512_test::test_sha512_size_one (gas: ~5697)
> [PASS] alexandria_math::tests::sha512_test::test_sha512_size_one (gas: ~633)
<         steps: 735926
>         steps: 85256
<         memory holes: 73172
>         memory holes: 4759
<         builtins: ("range_check_builtin": 142425, "bitwise_builtin": 2025)
>         builtins: ("range_check_builtin": 15812, "bitwise_builtin": 2720)

< [PASS] alexandria_math::tests::sha512_test::test_sha512_lorem_ipsum (gas: ~16591)
> [PASS] alexandria_math::tests::sha512_test::test_sha512_lorem_ipsum (gas: ~1823)
<         steps: 2149930
>         steps: 242168
<         memory holes: 217632
>         memory holes: 13335
<         builtins: ("range_check_builtin": 414769, "bitwise_builtin": 5929)
>         builtins: ("range_check_builtin": 45564, "bitwise_builtin": 8096)
```

- [ ] Yes
- [x] No

### DIFF
--- a/src/math/src/sha512.cairo
+++ b/src/math/src/sha512.cairo
@@ -255,16 +255,17 @@ fn from_u8Array_to_WordArray(data: Array<u8>) -> Array<Word64> {
     let mut i = 0;
 
     // Use precomputed powers of 2 for shift left to avoid recomputation
+    // Safe to use u64 coz we shift u8 to the left by max 56 bits in u64
     while (i < data.len()) {
-        let new_word: u128 = math_shl_precomputed::<u128>((*data[i + 0]).into(), TWO_56.into())
-            + math_shl_precomputed((*data[i + 1]).into(), TWO_48.into())
-            + math_shl_precomputed((*data[i + 2]).into(), TWO_40.into())
-            + math_shl_precomputed((*data[i + 3]).into(), TWO_32.into())
-            + math_shl_precomputed((*data[i + 4]).into(), TWO_24.into())
-            + math_shl_precomputed((*data[i + 5]).into(), TWO_16.into())
-            + math_shl_precomputed((*data[i + 6]).into(), TWO_8.into())
-            + math_shl_precomputed((*data[i + 7]).into(), TWO_0.into());
-        new_arr.append(Word64 { data: new_word.try_into().unwrap() });
+        let new_word: u64 = math_shl_precomputed::<u64>((*data[i + 0]).into(), TWO_56)
+            + math_shl_precomputed((*data[i + 1]).into(), TWO_48)
+            + math_shl_precomputed((*data[i + 2]).into(), TWO_40)
+            + math_shl_precomputed((*data[i + 3]).into(), TWO_32)
+            + math_shl_precomputed((*data[i + 4]).into(), TWO_24)
+            + math_shl_precomputed((*data[i + 5]).into(), TWO_16)
+            + math_shl_precomputed((*data[i + 6]).into(), TWO_8)
+            + math_shl_precomputed((*data[i + 7]).into(), TWO_0);
+        new_arr.append(Word64 { data: new_word });
         i += 8;
     };
     new_arr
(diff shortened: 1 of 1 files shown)


shramee:
Added comments in code to describe optimisations,

Mostly all them bit shifts and rot functions now use precomputed powers of two instead of trying to compute it on every steps. That gives us 9x performance gain.

src/math/src/sha512.cairo
```diff
@@ -8,6 +8,23 @@ pub const SHA512_LEN: usize = 64;
 
 pub const U64_BIT_NUM: u64 = 64;
 
+// Powers of two to avoid recomputing
+pub const TWO_56: u64 = 0x100000000000000;
```
gaetbout, line 12: why not u128 and avoid .into() 
(even though it is a free op ^^')
Same for some other consts

shramee: Because we also use it as u64, and this allows avoiding u128 `try_into()` u64

shramee: image

shramee: I had a couple `two_pow` calls which were using this before I cached all known powers of 2.
It'll be useful if we wanna use two_pow/shr (shl can be risky) somewhere with u64.

gaetbout: Some are only used as u128, but yeah, as it is free op, nvm :)

src/math/src/sha512.cairo
```diff
@@ -8,6 +8,23 @@ pub const SHA512_LEN: usize = 64;
...
+pub const TWO_48: u64 = 0x1000000000000;
+pub const TWO_40: u64 = 0x10000000000;
+pub const TWO_32: u64 = 0x100000000;
+pub const TWO_24: u64 = 0x1000000;
+pub const TWO_16: u64 = 0x10000;
+pub const TWO_8: u64 = 0x100;
+pub const TWO_4: u64 = 0x10;
+pub const TWO_2: u64 = 0x4;
+pub const TWO_1: u64 = 0x2;
+pub const TWO_0: u64 = 0x1;
```
tdelabro, line 22: Are those names descriptive enough?
Shouldn't `TWO_POW_<n>` be better?

As they are all public it is important

src/math/src/sha512.cairo
```diff
@@ -8,6 +8,23 @@ pub const SHA512_LEN: usize = 64;
...
+
+// Max u8 and u64 for bitwise operations
+pub const MAX_U8: u64 = 0xff;
+pub const MAX_U64: u128 = 0xffffffffffffffff;
```
tdelabro, line 26: Can't you use `core::interger::BoundedU8::max` rather than defining those again?
they are always inlined so it shouldn't have a cost.

tdelabro (changes requested):

src/math/src/sha512.cairo
tdelabro: Oh it looks like it was using Bounded::max before and you change it to this. Can you explain the perf improvement from it?

shramee: No perf improvement, just *aesthetics*.
Using that was breaking up some lines of code.
With smaller yet descriptive vars it stayed pretty.

src/math/src/sha512.cairo
shramee: Yes, totally.
Fixed!

shramee:
Actually there are more optimisations possible. But I’ll look at that another day.

tdelabro (approved):

PR #321, opened 2024-07-24
[DRAFT] Upgrade to 2.7.0-rc3

gaetbout:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Build-related changes

## What is the new behavior?

There might be some changes in the API. Should we also deprecate some stuff or can we just force to upgrade by removing them?


gaetbout:
@0xLucqs Ready for review. 
An upgrade to latest edition should still be performed, will open an issue

0xLucqs:
flemme de review je close bonne journée

packages/ascii/src/tests/test_ascii_integer.cairo
```diff
@@ -1,99 +1,99 @@
...
     // max u256 int in cairo is GitHub Copilot: The maximum u256 number in Cairo is `
     // 115792089237316195423570985008687907853269984665640564039457584007913129639935`.
-    let num: u256 = BoundedInt::max();
+    let num: u256 = Bounded::MAX;
     let ascii: Array<felt252> = num.to_ascii();
-    assert_eq!(ascii.len(), 3, "max u256 wrong len");
-    assert_eq!(*ascii.at(0), '1157920892373161954235709850086', "max u256 wrong first felt");
-    assert_eq!(*ascii.at(1), '8790785326998466564056403945758', "max u256 wrong second felt");
-    assert_eq!(*ascii.at(2), '4007913129639935', "max u256 wrong third felt");
+    assert_eq!(ascii.len(), 3);
```
0xLucqs, line 12: why ?

packages/data_structures/src/byte_reader.cairo
```diff
@@ -562,19 +576,10 @@ impl ByteReaderLenImpl<T, +Len<T>> of Len<ByteReaderState<T>> {
...
-    #[inline(always)]
-    fn index(self: @ByteArray, index: usize) -> @u8 {
-        @self.at(index).expect('Index out of bounds')
-    }
-}
-
-impl ByteReaderIndexViewImpl<
-    T, impl TIndexView: IndexView<T, usize, @u8>
-> of IndexView<ByteReaderState<T>, usize, @u8> {
+impl IntoU8Impl of Into<u8, @u8> {
```
0xLucqs, line 579: what am i watching here ? is this needed for some trait ?

gaetbout: this is the one thing I'm a bit puzzled, and had a lot pf back and forth with Ariel on Discord.
This is the only way I found to have this file working.
If you'd be kind enough to look into it, that'd be nice...

packages/data_structures/src/byte_reader.cairo
```diff
@@ -161,7 +171,10 @@ pub trait ByteReader<T> {
     fn len(self: @ByteReaderState<T>) -> usize;
 }
 
-impl ByteReaderImpl<T, +Drop<T>, +Len<T>, +IndexView<T, usize, @u8>> of ByteReader<T> {
+// use core::ops::index::IndexView;
+impl ByteReaderImpl<
+    T, +Drop<T>, +Len<T>, impl IndexViewImpl: IndexView<T, usize>, +Into<IndexViewImpl::Target, @u8>
```
0xLucqs, line 176:
```suggestion
    T, +Drop<T>, +Len<T>, +IndexView<T, usize>, +Into<IndexView::Target, @u8>
```

0xLucqs (changes requested):

0xLucqs (approved):

milancermak (commented):
Random things I've noticed (not necessarily enabled by 2.7.0):
* in `bytes/src/utils.cairo`, `uint_min` can be removed, replaced by `core::cmp::min`
* there's probably something that can be done with iterators on ByteReader, Queue, Stack, Vec and elsewhere(?)


Since we now have const arrays (❤️ `dw` ❤️), IMO it's worth adding optimized `pow2` and `pow10` as these are the most commonly used power functions. Even in Alexandria, all the bitshift fns use `pow(2, n)`.

packages/bytes/src/storage.cairo
```diff
@@ -131,16 +133,16 @@ fn inner_write_bytes(
...
-            Result::Ok(x) => x,
-            Result::Err(_) => {
-                // After writing 256 `uint128`s `index_in_chunk` will overflow and we move to the
-                // next chunk.
-                chunk += 1;
-                chunk_base = inner_bytes_pointer(address, chunk);
-                0
-            },
-        };
+        let (tmp, did_overflow) = index_in_chunk.overflowing_add(1);
```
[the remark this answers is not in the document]
gaetbout: Doesn't work :/

packages/math/src/extended_euclidean_algorithm.cairo
```diff
@@ -31,8 +31,9 @@ pub fn extended_euclidean_algorithm(a: u128, b: u128) -> (u128, u128, u128) {
 /// Update the step of the extended Euclidean algorithm.
 fn update_step(ref a: u128, ref old_a: u128, quotient: u128) {
     let temp = a;
-    let (bottom, _) = u128_overflowing_mul(quotient, temp);
-    a = u128_wrapping_sub(old_a, bottom);
+    let (bottom, _) = quotient.overflowing_mul(temp);
+    let (a_tmp, _) = old_a.overflowing_sub(bottom);
```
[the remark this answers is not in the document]
gaetbout: Doesn't work :/

packages/linalg/src/kron.cairo
```diff
@@ -22,14 +22,12 @@ pub fn kron<T, +Mul<T>, +Copy<T>, +Drop<T>,>(
...
-    while let Option::Some(x_value) = xs
-        .pop_front() {
-            let mut ys_clone = ys;
-            while let Option::Some(y_value) = ys_clone
-                .pop_front() {
-                    array.append(*x_value * *y_value);
-                }
-        };
+    for x_value in xs {
+        let mut ys_clone = ys;
```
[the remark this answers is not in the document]
gaetbout: needed otherwise the span is consumed and on the 2nd loop it fails

gaetbout:
> Random things I've noticed (not necessarily enabled by 2.7.0):
> 
> * in `bytes/src/utils.cairo`, `uint_min` can be removed, replaced by `core::cmp::min`
> * there's probably something that can be done with iterators on ByteReader, Queue, Stack, Vec and elsewhere(?)
> 
> Since we now have const arrays (❤️ `dw` ❤️), IMO it's worth adding optimized `pow2` and `pow10` as these are the most commonly used power functions. Even in Alexandria, all the bitshift fns use `pow(2, n)`.

updated utils.cairo
y, but I'd leave it for another PR as those are "extra" changes and this PR is big enough

Also would leave for another PR 😅 , please open an issue for that

packages/linalg/src/kron.cairo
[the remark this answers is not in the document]
milancermak: Tests pass however 😱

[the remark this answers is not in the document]
gaetbout: lemme double check, maybe I messed up somewhere else while testing :D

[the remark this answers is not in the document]
gaetbout: Actually running in a lowering crash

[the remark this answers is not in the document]
milancermak: Got the crash as well, submitted an issue to Scarb. Might be worth just removing the `mut` for now?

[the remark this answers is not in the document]
gaetbout: Done

enitrat:
@gaetbout we should suppress the warnings on compilation otherwise it's going to appear on all dependent repositories

gaetbout:
I removed all warnings that are from the lib. 
I believe the new warnings are from using now deprecated Alexandria stuff

packages/math/src/u512_arithmetics.cairo
```diff
@@ -1,18 +1,23 @@
...
 use core::result::ResultTrait;
 
-pub fn u256_overflow_add(lhs: u256, rhs: u256) -> Result<u256, u256> implicits(RangeCheck) nopanic {
-    let (sum, overflow) = core::integer::u256_overflowing_add(lhs, rhs);
+#[deprecated(
+    feature: "deprecated-u256_overflow_add",
+    note: "Use `use core::num::traits::OverflowingAdd`.",
+    since: "2.7.0"
+)]
+pub fn u256_overflow_add(lhs: u256, rhs: u256) -> Result<u256, u256> implicits(RangeCheck) {
```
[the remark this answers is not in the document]
gaetbout: tbh we could remove all the code that is now deprecated, but to leave some time to the ppl that might be using it, maybe better to add a warning until next version

packages/bytes/src/bytes.cairo
```diff
@@ -4,8 +4,8 @@ use alexandria_bytes::utils::{
 use alexandria_math::sha256::sha256;
```
enitrat, line 4: should be replaced in favor of syscall

packages/bytes/src/bytes.cairo
```diff
@@ -116,8 +117,18 @@ pub trait BytesTrait {
...
     fn concat(ref self: Bytes, other: @Bytes);
     /// keccak hash
+    #[deprecated(
+        feature: "deprecated-keccak",
+        note: "Use `core::keccak::compute_keccak_byte_array`.",
+        since: "2.7.0"
+    )]
     fn keccak(self: @Bytes) -> u256;
     /// sha256 hash
+    #[deprecated(
```
enitrat, line 127: shouldnt be deprecated. `Bytes` != `ByteArray` - and if we consider this deprecated then the whole Bytes type should be deprecated.

packages/bytes/src/bytes.cairo
```diff
@@ -116,8 +117,18 @@ pub trait BytesTrait {
     /// concat with other Bytes
     fn concat(ref self: Bytes, other: @Bytes);
     /// keccak hash
+    #[deprecated(
```
enitrat, line 120: same as below

packages/bytes/src/bytes.cairo
enitrat: `sha256(self: @Bytes)` should use the `sha256` syscall

packages/bytes/src/storage.cairo
```diff
@@ -86,16 +87,17 @@ fn inner_read_bytes(address_domain: u32, address: StorageAddress) -> SyscallResu
...
-            Result::Err(_) => {
-                // After reading 256 `uint128`s `index_in_chunk` will overflow and we move to the
-                // next chunk.
-                chunk += 1;
-                chunk_base = inner_bytes_pointer(address, chunk);
-                0
-            },
-        };
+
+        let (tmp, did_overflow) = index_in_chunk.overflowing_add(1);
```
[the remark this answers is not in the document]
enitrat: using `checked_add` instead of overflowing_add would solve the problem. same for all subsequent cases

packages/bytes/src/bytes.cairo
enitrat: here this shouldn't be imported in the `bytes` module. the syscall should be imported here and used for sha256(@Bytes)

packages/bytes/src/bytes.cairo
gaetbout: Anyway since it is using a deprecated fn (math::sha256) the deprecate will propagate upwards.
Maybe we can make this whole bytes stuff deprecated indeed

packages/math/src/u512_arithmetics.cairo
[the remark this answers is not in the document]
enitrat: yeah but this has been wrongly been made public in the first place :/

enitrat:
my point is: what currently is used inside alexandria itself should not be triggering warnings

packages/bytes/src/storage.cairo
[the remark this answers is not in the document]
gaetbout: Fixed where it could be done in e8546cf

packages/math/src/u512_arithmetics.cairo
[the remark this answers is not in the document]
gaetbout: @0xLucqs Opinion?

packages/bytes/src/bytes.cairo
gaetbout: solved at 846621a1b731590cf316f9bec20e75cfe39b88a4, lmk if good :)

packages/bytes/src/bytes.cairo
gaetbout: Should we make the whole Byte deprecated?

packages/math/src/u512_arithmetics.cairo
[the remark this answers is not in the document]
gaetbout: I updated it

gaetbout:
@0xLucqs ?

enitrat (approved):

gaetbout:
@0xLucqs ? :D

0xLucqs:
I want @milancermak to approve as well

packages/merkle_tree/src/merkle_tree.cairo
```diff
@@ -104,18 +106,18 @@ pub impl MerkleTreeImpl<T, +HasherTrait<T>, +Copy<T>, +Drop<T>> of MerkleTreeTra
...
-                // We need to check if the current node is smaller than the current element of the proof.
-                // If it is, we need to swap the order of the hash.
-                current_node =
-                    if Into::<felt252, u256>::into(current_node) < (*proof_element).into() {
-                        self.hasher.hash(current_node, *proof_element)
-                    } else {
-                        self.hasher.hash(*proof_element, current_node)
-                    }
-            };
+        while let Option::Some(proof_element) = proof.pop_front() {
```
milancermak, line 109: Can be modernized as `for proof_element in proof`

packages/merkle_tree/src/merkle_tree.cairo
```diff
@@ -129,18 +131,18 @@ pub impl MerkleTreeImpl<T, +HasherTrait<T>, +Copy<T>, +Drop<T>> of MerkleTreeTra
...
-                // We need to check if the current node is smaller than the current element of the proof.
-                // If it is, we need to swap the order of the hash.
-                leaf =
-                    if Into::<felt252, u256>::into(leaf) < (*proof_element).into() {
-                        self.hasher.hash(leaf, *proof_element)
-                    } else {
-                        self.hasher.hash(*proof_element, leaf)
-                    };
-            };
+        while let Option::Some(proof_element) = proof.pop_front() {
```
milancermak, line 134: Same as above, `for proof_element in proof`

milancermak (approved):
LGTM! Leaving just a bunch of nits.

If you fancy, there's a bunch of places where the `while-let` can be updated to a `for-in` (e.g. in array_ext.cairo or sha256.cairo, though the latter is deprecated, so meh).

packages/numeric/src/trapezoidal_rule.cairo
```diff
@@ -1,6 +1,6 @@
 use core::array::SpanTrait;
 use core::num::traits::Zero;
-use core::option::OptionTrait;
+use core::ops::{MulAssign, AddAssign};
```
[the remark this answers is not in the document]
gaetbout: Removed

packages/merkle_tree/src/merkle_tree.cairo
gaetbout: Done

packages/merkle_tree/src/merkle_tree.cairo
gaetbout: Done

PR #340, opened 2024-12-10
Fix read_bytes31

augustbleeds:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

read_bytes31 does not begin at the offset. This is incorrect standalone behavior and creates an infinite loop in the .into() function (from ByteTrait to ByteArray) if there is more than 61 bytes in the bytes struct

Issue Number: N/A

## What is the new behavior?

Starts at the right offset

- [ ] Yes
- [x] No

### DIFF
--- a/packages/bytes/src/tests/test_bytes.cairo
+++ b/packages/bytes/src/tests/test_bytes.cairo
@@ -688,7 +688,7 @@ fn test_bytes_sha256() {
 #[test]
 fn test_byte_array_conversions() {
     let bytes = BytesTrait::new(
-        52,
+        64,
         array![
             0x01020304050607080910111213141516,
             0x16151413121110090807060504030201,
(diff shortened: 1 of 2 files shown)


packages/bytes/src/tests/test_bytes.cairo
```diff
@@ -688,7 +688,7 @@ fn test_bytes_sha256() {
 #[test]
 fn test_byte_array_conversions() {
     let bytes = BytesTrait::new(
-        52,
+        64,
```
augustbleeds, line 691: a large enough bytes value will make the prior code fail

gaetbout:
Gm, thanks for the PR,
Could you add a test that fails with the previous behaviour? :)

augustbleeds:
> Gm, thanks for the PR, Could you add a test that fails with the previous behaviour? :)

Sounds good 👍 , I modified the read_bytes31 and the conversion function to fail under previous behavior

gaetbout (approved):

PR #360, opened 2025-04-14
feat: Add disclaimer for unaudited code

edisontim:
This adds a disclaimer for the ed25519 sig verification. The code probably isn't safe and shouldn't be used in a production environment for sensitive operations


ermvrs:
We are about to fix a bug on ed25519 rn. But you re right. we will add a mention to all unaudited signature verification methods.

edisontim:
> We are about to fix a bug on ed25519 rn. But you re right. we will add a mention to all unaudited signature verification methods.

Will close this PR then

